PR #1, opened 2024-08-02
feat: migrate tox tooling to the monorepo

jedel1043:
This adds several changes that make it really easy to format, lint, and test against some or all of the charms in the repository:

- Adds a `repository.py` file that takes care of all the automation.
- Extracts all common libraries to an `external` directory.
- Migrates the integration tests to the root.
- Rewrites our tox file to forward to `repository.py`.

In general, this makes it so that we can still run everything using `tox -e command`, and `repository.py` takes care of setting up everything for us.


tox.ini
NucciTheBoss, line 1: This file has inconsistent indentation. Some sections use 2 spaces while others use 4 spaces. Typically I try to use 4 spaces for tox files.

tox.ini
```diff
@@ -0,0 +1,94 @@
+# Global tox file
+
+# This file is used to invoke tox in individual charms
```
NucciTheBoss, line 3: [nit]: I don't think these two comments are necessary. I think it's clear based on the file's location within the repository that it is the root tox file.

tox.ini
```diff
@@ -0,0 +1,94 @@
...
+[testenv]
+allowlist_externals =
+  {toxinidir}/repository.py
+
+[testenv:fetch-lib]
+description = Fetch or bump the specified library.
+basepython = python3
+deps = pyyaml
+commands =
+  python3 {toxinidir}/repository.py -v fetch-lib {posargs}
```
NucciTheBoss, line 24: [praise]: This is great to have!

tox.ini
```diff
@@ -0,0 +1,94 @@
+# Global tox file
+
+# This file is used to invoke tox in individual charms
+
+[tox]
+skipsdist = True
+envlist = pep8,py3
```
NucciTheBoss, line 7: This isn't the correct usage of `envlist` here. `envlist` is used to declare the default environments that will be run if the user does not pass any environment as a flag to `tox run`: https://tox.wiki/en/latest/config.html#env_list

`fmt`, `lint`, `type`, and `unit` can be in the default envlist since they are not beefy like the integration tests.

```suggestion
envlist = fmt,lint,type,unit
```

tox.ini
```diff
@@ -0,0 +1,94 @@
...
+sitepackages = False
+skip_missing_interpreters = False
+minversion = 3.18.0
```
NucciTheBoss, line 10: Do we want this to be tox v4 instead? Let me know your thoughts.

.stestr.conf
NucciTheBoss, line 1: Where is `stestr` used here? Or is it something called externally not by tox? I see it mentioned in `repository.py`

repository.py
```diff
@@ -0,0 +1,543 @@
...
+
+import yaml
+
+ROOT_DIR = pathlib.Path(__file__).parent
+EXTERNAL_LIB_DIR = ROOT_DIR / "external" / "lib"
+BUILD_PATH = ROOT_DIR / ".slurm-build"
+BUILD_FILE = ".slurm-build.yaml"
+UTILITY_FILES = [
+    ROOT_DIR / ".stestr.conf",
+    ROOT_DIR / ".jujuignore",
```
NucciTheBoss, line 22: Do we need a `.jujuignore` file here?

repository.py
```diff
@@ -0,0 +1,543 @@
...
+import subprocess
+import os
+import glob
+
+import yaml
+
+ROOT_DIR = pathlib.Path(__file__).parent
+EXTERNAL_LIB_DIR = ROOT_DIR / "external" / "lib"
+BUILD_PATH = ROOT_DIR / ".slurm-build"
+BUILD_FILE = ".slurm-build.yaml"
```
NucciTheBoss, line 19: Hmm... I'm not sure how I feel about a hidden .build file that will dictate how the charm is built. Typically I like build files to be explicit. What are your thoughts on this?

Also, why `.slurm-build.yaml` instead of just `.build.yaml`? Not sure we need the additional namespacing here on the file if repository.py is a general tool that can be used by charm monorepos.

repository.py
```diff
@@ -0,0 +1,543 @@
...
+import shutil
+import subprocess
+import os
+import glob
+
+import yaml
+
+ROOT_DIR = pathlib.Path(__file__).parent
+EXTERNAL_LIB_DIR = ROOT_DIR / "external" / "lib"
+BUILD_PATH = ROOT_DIR / ".slurm-build"
```
NucciTheBoss, line 18: See my earlier comment about using `_build` instead. Curious to know your thoughts on this.

repository.py
```diff
@@ -0,0 +1,543 @@
+#!/usr/bin/env python
+
+# Taken from https://opendev.org/openstack/sunbeam-charms/src/commit/5b37e0a6919668f23b8c7b148717714889fd4381/repository.py
+
+import logging
+import argparse
+import dataclasses
```
NucciTheBoss, line 7: Why not just `from dataclasses import dataclass`? Looks like we're only using the `dataclass` decorator here.

repository.py
```diff
@@ -0,0 +1,543 @@
...
+import pathlib
+import shutil
+import subprocess
+import os
+import glob
```
NucciTheBoss, line 12: I think these imports need to be sorted by ruff/isort.

tests/integration/conftest.py
NucciTheBoss, line 1: Should we just assume that the Slurm charms will always be built from source or not worry about separate charms? Removes the additional logic required to pull from Charmhub.

repository.py
```diff
@@ -0,0 +1,543 @@
...
+
+import yaml
+
+ROOT_DIR = pathlib.Path(__file__).parent
+EXTERNAL_LIB_DIR = ROOT_DIR / "external" / "lib"
+BUILD_PATH = ROOT_DIR / ".slurm-build"
+BUILD_FILE = ".slurm-build.yaml"
+UTILITY_FILES = [
+    ROOT_DIR / ".stestr.conf",
```
NucciTheBoss, line 21: Are we using `stestr` at all? Could be useful

NucciTheBoss (changes requested):
_repository.py_ will be great to have for automating various charm actions. Less script-kiddying required than if we were to use `make` or `just`.

I left several comments since this is a bigger PR, but overall I'm happy with how the monorepo is shaping up!

.stestr.conf
jedel1043: Remnants from the original tooling. I'll delete it but we can consider using `stestr` in the future.

repository.py
jedel1043: I don't want to conflict with some other tool that also uses `build.yaml` for its config. Agreed with using a non-hidden file though.

tox.ini
jedel1043: Fixed.

.gitignore
```diff
@@ -8,4 +8,5 @@ __pycache__/
 .idea
 .vscode/
 version
-.ruff_cache
\ No newline at end of file
+.ruff_cache/
+.slurm-build/
```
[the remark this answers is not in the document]
jedel1043: Used the `build` directory instead.

tox.ini
jedel1043: Removed them.

repository.py
```diff
@@ -0,0 +1,552 @@
...
+import shutil
+import subprocess
+from dataclasses import dataclass
+
+import yaml
+
+ROOT_DIR = pathlib.Path(__file__).parent
+EXTERNAL_LIB_DIR = ROOT_DIR / "external" / "lib"
+BUILD_PATH = ROOT_DIR / "build"
+BUILD_FILE = "slurm-build.yaml"
```
NucciTheBoss, line 21: Tbh I don't think we need to worry about conflicting with other build systems here if we were to use the name `build.yaml`. Quick search seems to indicate the Dart is the only major language/ecosystem that uses a `build.yaml` file to control project builds. I'd have some questions if we had a Dart project in this repo 😶‍🌫️

If we don't want to have a generic `build.yaml` file, how about we use the name `charm-build.yaml` instead? I'm concerned that someone looking from the outside in at the Slurm charms would think that `slurm-build.yaml` dictates how we're building Slurm inside of the charms rather than dictating what charm libraries we're linking in as part of the charm packing process.

NucciTheBoss (changes requested):
Looks good! Just two small comments I'd like you to address and then I'm ready to merge this PR if it's ready.

repository.py
```diff
@@ -0,0 +1,552 @@
...
+import pathlib
+import shutil
+import subprocess
+from dataclasses import dataclass
+
+import yaml
+
+ROOT_DIR = pathlib.Path(__file__).parent
+EXTERNAL_LIB_DIR = ROOT_DIR / "external" / "lib"
+BUILD_PATH = ROOT_DIR / "build"
```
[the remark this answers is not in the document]
jedel1043: changed to _build

repository.py
jedel1043: changed to build.yaml

NucciTheBoss (approved):
LGTM! :shipit:

PR #4, opened 2024-08-06
ci: enable CI actions

jedel1043:
This adds another command in tox to generate a Charmhub token, which should be useful to regenerate it in an easier way.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,0 +1,70 @@
+# Copyright 2024 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Bug Report
+description: File a bug report
+labels: ["Type: Bug", "Status: Triage"]
+body:
+  - type: markdown
+    attributes:
+      value: >
+        Thanks for taking the time to fill out this bug report! Before submitting your issue, please make
+        sure you are using the latest version of the charm. If not, please switch to the latest version of this charm
+        before posting your report to make sure it's not already solved.
+  - type: textarea
+    id: bug-description
+    attributes:
+      label: Bug Description
+      description: >
+        Provide a description of the issue you are facing. If applicable, add screenshots to help explain the problem.
+    validations:
+      required: true
+  - type: textarea
+    id: reproduction
+    attributes:
+      label: To Reproduce
+      description: >
+        Please provide a step-by-step instruction of how to reproduce the behaviour.
+      placeholder: |
+        1. `juju deploy ...`
+        2. `juju integrate ...`
+        3. `juju status --integrations`
+    validations:
+      required: true
+  - type: textarea
+    id: environment
+    attributes:
+      label: Environment
+      description: >
+        We need to know a bit more about the context in which you run the charm.
+        - Are you running Juju locally, on lxd, in multipass or on some other platform?
+        - What track and channel you deployed the charm from (ie. `latest/edge` or similar).
+        - Version of any applicable components, like the juju snap, the model controller, lxd, microk8s, and/or multipass.
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: >
+        Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
+        Fetch the logs using `juju debug-log --replay`. Additional details on how to retrieve logs are available in the juju
+        documentation at https://juju.is/docs/juju/log.
+      render: shell
+    validations:
+      required: true
+  - type: textarea
+    id: additional-context
+    attributes:
+      label: Additional context
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,91 @@
+# Copyright 2024 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


jedel1043:
Looking good so far. Let's see if integration tests work.

charms/slurmctld/charmcraft.yaml
NucciTheBoss, line 1: [foresight]: we'll want to update this metadata in another PR. Just have everything come through the monorepo issue tracker.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,70 @@
...
+      required: true
+  - type: textarea
+    id: reproduction
+    attributes:
+      label: To Reproduce
+      description: >
+        Please provide a step-by-step instruction of how to reproduce the behaviour.
+      placeholder: |
+        1. `juju deploy ...`
+        2. `juju relate ...`
```
NucciTheBoss, line 41: Should update to reflect Juju 3 terminology:

```suggestion
        2. `juju integrate ...`
```

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,70 @@
...
+        3. `juju status --relations`
```
NucciTheBoss, line 42:
```suggestion
        3. `juju status --integrations`
```

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,70 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: >
+        Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
+        Fetch the logs using `juju debug-log --replay`. Additional details on how to retrieve logs are available in the juju
+        documentation at https://juju.is/docs/olm/juju-logs.
```
NucciTheBoss, line 63: Juju isn't referred to as the "operator lifecycle manager" anymore. The `olm` slug doesn't exist anymore on the Juju documentation :sweat_smile: 

```suggestion
        documentation at https://juju.is/docs/juju/log.
```

.github/PULL_REQUEST_TEMPLATE.md
NucciTheBoss, line 1: Let's hold off on adding a PR template right now. I think this template needs to be reworked. The checklist doesn't always get updated correctly, _Description_ header isn't necessary, and code should always be tested in reproducible way.

Better to have wider discussion about what the template should look like in a separate PR.

repository.py
```diff
@@ -331,6 +334,25 @@ def stage_cli(
...
+        "login",
+        "--export=slurm-charms.auth"
+    ] + [f"--charm={charm}" for charm in slurm_charms] + [
+        "--permission=package-manage-metadata",
+        "--permission=package-manage-releases",
+        "--permission=package-manage-revisions",
+        "--permission=package-view-metadata",
+        "--permission=package-view-releases",
+        "--permission=package-view-revisions",
+        "--ttl=7776000"
```
NucciTheBoss, line 353: How do we feel about adding the comment 90 days so we have a better idea when the token is going to expire :sweat_smile: 

```suggestion
        "--ttl=7776000"  # 90 days
```

.gitignore
```diff
@@ -10,3 +10,4 @@ __pycache__/
 version
 .ruff_cache/
 _build
+slurm-charms.auth
```
NucciTheBoss, line 13: What if we used a secret dotfile named `.secret` or `.charmhub.secret`? We could then configure GitHub to scream at us if this file were to be accidentally committed and pushed to the public repo. Also, I think "secret" slightly communicates better that this sensitive information that we don't want to expose.

Also [nit]: missing newline at the end of the file :sweat_smile: 

```suggestion
slurm-charms.auth

```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,88 @@
...
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: python3 -m pip install tox
+      - name: Run linters
+        run: tox -e lint
```
NucciTheBoss, line 42: We should have the syntax consistent here. IIRC tox v4  prefers `tox run -e` rather than just `tox -e ...`:

```suggestion
        run: tox run -e lint
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,88 @@
...
+  type-check:
+    name: Static type checking
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: python3 -m pip install tox
+      - name: Run tests
+        run: tox -e type
```
NucciTheBoss, line 64:
```suggestion
        run: tox run -e type
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,88 @@
...
+  unit-test:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: python3 -m pip install tox
+      - name: Run tests
+        run: tox -e unit
```
NucciTheBoss, line 53:
```suggestion
        run: tox run -e unit
```

NucciTheBoss (changes requested):
Generating the token will be super helpful. Don't need to clunk our way around the CLI.

Just some comments, but overall I think this PR looks good!

NucciTheBoss:
Hmm... might need to run the integration tests locally to get a better idea of what the issues. Seems like `slurmctld` is failing to reach an active state which indicates that something is going wrong with the setup process.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,91 @@
...
+# limitations under the License.
+
+name: slurm charms tests
+on:
+  workflow_call:
+  pull_request:
+    branches:
+      - main
+      - experimental
+
```
NucciTheBoss, line 22: [nit]: unnecessary newline here. Typically `yamllint` will complain about this.

```suggestion
```

NucciTheBoss (approved):
Looks good! Really tiny change needed in _ci.yaml_, but I can commit the fix as a suggestion. Just removing a unnecessary additional newline after the "on" section.

PR #7, opened 2024-08-07
fix: make the tox fetch-lib command work

jedel1043:
Apparently `fetch-lib` needs at least the `charmcraft.yaml` file to work.


NucciTheBoss (approved):
LGTM!

One could wonder if we could just assemble this file on the fly when we invoke the `fetch-lib` action and then clean it up after we're done. Probably more trouble than it's worth however :sweat_smile:

PR #12, opened 2024-08-07
test: add additional integration checks

jedel1043:
Combines all the checks from the integration tests of all the charms.

### DIFF
--- a/repository.py
+++ b/repository.py
@@ -419,6 +419,7 @@ def fmt_cli(
 ):
     """Apply correct formatting to code."""
     files = get_source_dirs(slurm_charms)
+    files.append(str(ROOT_DIR / "tests"))
     logging.info(f"Running black for directories {files}")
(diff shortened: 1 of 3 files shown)


NucciTheBoss (approved):
LGTM! This will be nice to have when making changes to all the charms!

One thing we should consider in the future is compiling a list of tests we want to have for the Slurm charms. For example, I could see us having a test for cycling out the munge key, validating the ket, updating slurm.conf configuration, etc. Out of scope for this PR, but eventually we should look into building out the test suite to validate ALL THE THINGS™️

NucciTheBoss:
@jedel1043 I don't think we want to disallow base branch updates?

image

NucciTheBoss:
I needed to untick this restriction so that I can merge this PR. We should figure out what branch protections we want to set in place. If we do want to disallow base branch updates, we should at least add the @charmed-hpc/charmed-hpc-core-devs have bypass privileges for that specific rule but still require that CI succeeds. That way maintainers who aren't considered core devs can't mess with the base branch.

image

PR #21, opened 2024-08-20
chore(slurmctld): use peer relation for ingress ip

jamesbeedy:
These changes add a peer relation for the slurmctld charm and replace using the slurmd interface to obtain the ingress_address with the new slurmctld-peer relation.

The reason for this change is that we do not want to depend on the existence of the slurmd relation in order to know our ip.

Using a peer relation we will always have resolvability so long as juju knows the ip address of the unit.


NucciTheBoss (changes requested):
Overall, looks great! Just a couple of suggested changes for the custom_exceptions module.

This is will great to have for when we go to enable HA support for the slurmctld operator!

PR #25, opened 2024-09-09
POC for integration with ROCm drivers

jedel1043:
The goal is to be able to automatically install ROCm drivers for compute nodes.


charms/slurmd/src/slurmd_ops.py
```diff
@@ -129,6 +129,47 @@ def install(self) -> bool:
...
+
+        rocm_repo = apt.DebianRepository(
+            enabled=True,
+            repotype="deb",
+            uri="https://repo.radeon.com/rocm/apt/6.2",
+            release=distro.codename(),
+            groups=["main"],
+            gpg_key_filename=self._KEYRING_PATH,
+            options={
+                "arch": "amd64"
```
wolsen, line 165: I know this is draft, but please make sure we're not hardcoding this arch string in the future. There are other architectures in the mix (e.g. amd64, riscv64, etc).

jedel1043: Definitely! I had a todo comment here on a previous version, but your review will be a good reminder to fix this.

jedel1043:
Closing this for now since the PR has gone stale and I suspect supporting ROCm drivers would have a very different shape in the future.

PR #27, opened 2024-09-24
fix: reintroduce NHC config template

jedel1043:
This fixes our CI pipeline.


charms/slurmd/src/templates/nhc.conf.tmpl
```diff
@@ -0,0 +1,170 @@
+# OSD enforces short hostnames
```
jamesbeedy, line 1:
```suggestion
# Enforce short hostnames to match the node names as tracked by slurm.
```

jamesbeedy (approved):
+1 niceeee

PR #28, opened 2024-10-01
feat: introduce TiCS analysis to the repo

jedel1043:
This introduces a Github action that uses the TiCS framework to send code quality statistics to TIOBE.

### DIFF
--- a/.github/workflows/nightly-TICS.yaml
+++ b/.github/workflows/nightly-TICS.yaml
@@ -1,0 +1,36 @@
+name: Nightly TICS scan
+
+on:
+  schedule:
+    - cron: '15 17 * * *' # Every day at 17:15 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  TICS:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          python3 -m pip install tox pylint flake8
+          python3 -m pip install --requirement test-requirements.txt
+
+      - name: Run coverage tests
+        run: |
+          tox -e unit
+
+      - name: TICS GitHub Action
+        uses: tiobe/tics-github-action@v3
+        with:
+          mode: qserver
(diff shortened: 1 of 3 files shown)


jedel1043:
Riiight, external branches don't have access to the secret. We'll leave it as a nightly run then.

.github/workflows/nightly-TICS.yaml
```diff
@@ -0,0 +1,36 @@
+name: Nightly TICS scan
+
+on:
+  schedule:
+    - cron: '44 1 * * *'
```
NucciTheBoss, line 5: [nit]: Can you comment the time of day that this action will run? Helpful for anyone coming into the project that might not be super familiar with cron syntax:

```suggestion
    - cron: '44 1 * * *' # Every day at 1:44 AM
```

Also, is there a specific reason why you chose for this action to run at 1:44 AM? Is it because everyone else chooses to have their cron job run at 12:00 AM on GitHub? :sweat_smile:

NucciTheBoss (changes requested):
Just a small nit, but other than that, this looks great! Super helpful to have deeper insight into the code quality of the Slurm charms :star_struck:

.github/workflows/nightly-TICS.yaml
jedel1043: Yeah it's just a random time that was taken from https://github.com/medvied/charm-ceph-mon/blob/master/.github/workflows/nightly-TICS.yaml.
It's not recommended to schedule jobs at "standard" times like 00:00 am or 12:00 pm:
> https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule
The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. If the load is sufficiently high enough, some queued jobs may be dropped. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.

NucciTheBoss:
Re-running CI. New action shouldn't affect the integration tests, so probably just an intermittent failure.

PR #34, opened 2024-10-23
Fix the resume action in the slurmctld charm

matheushent:
This PR fixes the *resume* action in the *slurmctld* charm. The source error can be seen on [this](https://app.warp.dev/block/MxGNPLvRGpmjegds477jJz) link.

Essentially, the command to resume the nodes has been patched from `scontrol update nodename=<names> state=resume` to `scontrol update nodename=<names> state=idle`.

As I understand [Slurm's documentation](https://slurm.schedmd.com/scontrol.html#OPT_RESUME), there's no operational change applied by this PR.

### DIFF
--- a/charms/slurmctld/src/charm.py
+++ b/charms/slurmctld/src/charm.py
@@ -204,7 +204,7 @@ class SlurmctldCharm(CharmBase):
         event.log(f"Resuming {nodes}.")
 
         try:
-            cmd = f"scontrol update nodename={nodes} state=resume"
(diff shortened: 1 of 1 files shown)


NucciTheBoss (approved):
LGTM :+1: 

Eventually we'll roll this into the `scontrol` method we've added to `slurm_ops` as all calls will be streamed to the juju debug log, but it'll be good to have the `resume` action good and going :star_struck:

PR #41, opened 2024-11-18
feat(slurmctld): enable `cos-agent` for metrics collection from Slurm cluster

NucciTheBoss:
This PR adds in the original COS-related patches from the `experimental` branch of the Slurm charms, but with an extra dose of Slurm back-end support provided by the `slurm_ops` charm library. This means that COS both now works with the Slurm snap and Slurm Debian package. 

This is a bit more difficult to write integration tests for as microk8s + LXD will quickly overwhelm your average GitHub runner, but the next steps are to create a COS-lite/Grafana Terraform module that can be used by `charmed-hpc-terraform` to deploy the Slurm charms with COS if desired. Here's what the default dashboard looks like currently on Grafana:

image

### Next steps

* Fix the publishing CI.
* Cache charm builds so they go faster in CI.
* Add grafana-agent to Charmed HPC's Terraform plan.
* Add more dashboards + understand what metrics folks want to be shown by default.
* Add more alert rules as we see fit/are requested by folks for the charms.
* Documentation for how to set up Slurm with COS.

### DIFF
--- a/charms/slurmctld/build.yaml
+++ b/charms/slurmctld/build.yaml
@@ -1,4 +1,5 @@
 external-libraries:
+  - charms.grafana_agent.v0.cos_agent
   - charms.hpc_libs.v0.is_container
   - charms.hpc_libs.v0.slurm_ops
   - charms.operator_libs_linux.v0.apt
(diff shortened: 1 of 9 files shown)


dsloanm (approved):
No comments - looks good!

NucciTheBoss:
> No comments - looks good!

Based :shipit: 

One thing will need to address in the future is that if we have multiple exporters on the same machine, we'll need to add support for configuring the port that the Slurm exporter uses. Right now though we can just focus on building out the documentation + deployment plan we need.

Issue #36, opened 2024-11-09
Sane defaults for MemSpecLimit and SelectTypeParameters


jamesbeedy:
### Bug Description

charmed-hpc should prevent jobs from exhausting all memory on the system by default.

1) Currently, job allocations can exhaust all available memory on the system. Set `MemSpecLimit=1024` to ensure systems always have 1G available memory that jobs cannot consume.

2) The slurm configuration provided by the slurm charms does not enforce memory limits on job allocations. Set `SelectTypeParameters=CR_CPU_Memory` to enable slurm + cgroup memory allocation enforcement. 

OPT_CR_CPU_Memory -> https://slurm.schedmd.com/slurm.conf.html#OPT_CR_CPU_Memory

MemSpecLimit -> https://slurm.schedmd.com/slurm.conf.html#OPT_MemSpecLimit

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

NucciTheBoss:
I agree that we should ensure that at least 1G is free on all the compute nodes so that we don't accidentally kill the daemons we need on the machines. _Nobody likes accidentally taking a node offline..._

We should also enforce memory limits on nodes to prevent there being noisy neighbors on compute nodes.

PR #46, opened 2024-11-24
enhancement(slurmctld): Make slurm respect memory constraints

jamesbeedy:
These changes introduce default values for `SelectTypeParameters` and `MemSpecLimit`.

`SelectTypeParameters=CR_CPU_Memory` to enable memory allocation enforcement.

`MemSpecLimit=1024` to ensure systems always have 1G available memory that jobs cannot consume.

Fixes: #36

### DIFF
--- a/charms/slurmd/src/charm.py
+++ b/charms/slurmd/src/charm.py
@@ -7,7 +7,6 @@
 import logging
 from typing import Any, Dict, cast
 
-from constants import CHARM_MAINTAINED_NODE_PARAMETERS
 from interface_slurmctld import Slurmctld, SlurmctldAvailableEvent
 from ops import (
     ActionEvent,
--- a/charms/slurmd/src/charm.py
+++ b/charms/slurmd/src/charm.py
@@ -7,6 +7,7 @@
 import logging
 from typing import Any, Dict, cast
 
+from constants import CHARM_MAINTAINED_NODE_PARAMETERS
 from interface_slurmctld import Slurmctld, SlurmctldAvailableEvent
 from ops import (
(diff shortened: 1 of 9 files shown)


charms/slurmd/src/constants.py
NucciTheBoss, line 1: Given that this constant is a simple dictionary that's only referenced once, it might be better to just refer to `{"MemSpecLimit": "1024"}` directly in the `get_node` method of `SlurmdCharm`. I could see us adding the constants module back in if we produce more constants for the slurmd operator.

charms/slurmd/src/charm.py
```diff
@@ -326,6 +327,7 @@ def get_node(self) -> Dict[Any, Any]:
         node = {
             "node_parameters": {
                 **machine.get_slurmd_info(),
+                **CHARM_MAINTAINED_NODE_PARAMETERS,
```
NucciTheBoss, line 330: Since `CHARMED_MAINTAINED_NODE_PARAMETERS` is only referenced once here, and only `MemSpecLimit` is set in the constant, perhaps it would be better to set `MemSpecLimit` directly here for now:

```suggestion
                "MemSpecLimit": "1024",
```

NucciTheBoss (changes requested):
Just one small comment I'd like addressed, and then I'm good to merge this PR!

NucciTheBoss (approved):
LGTM! :shipit:

NucciTheBoss:
There seems to be a rebase conflict preventing me rebasing and merging, so I'll squash and merge instead.

PR #42, opened 2024-11-22
fix(ci): do not use destructive packing when building charms for release

NucciTheBoss:
This PR sets the `destructive-mode` option for the `upload-charm` action to false. Pulling the down the base LXD instance will add some extra time onto the workflow run, but it make us not be dependent on the base OS of the GitHub runner VM to release the Slurm charms. This is especially helpful for it we need to release multiple the Slurm charms with support for multiple bases. Eventually we'll either want to replace the `cryptography` dependency that takes so long to build, or use the `ccc` tool to cache the Slurm charms dependencies so that the build for them is quicker.

Just for added insurance, I also updated the `CHARMCRAFT_AUTH` secret.

### Related issues

* Fixes #39

PR #47, opened 2024-11-25
fix(ci): initialize lxd before packing charm

NucciTheBoss:
The `upload-charm` action does not initialize LXD which will cause the worklflow to fail if destructive-mode is set to false.

This PR is a follow-on fix for https://github.com/charmed-hpc/slurm-charms/pull/42 since the `upload-charm` action doesn't set up LXD itself.


jedel1043 (approved):
Looks good! Hopefully this is enough to fix CI

PR #51, opened 2024-12-08
feat(all)!: upgrade Slurm charms to Noble

NucciTheBoss:
This PR upgrades the Slurm charms to Noble from Jammy. We get several benefits from this such as access to a newer base and package versions + more recent hardware support.

BREAKING CHANGES:

1. The Slurm charms no longer use the PPA `ppa:ubuntu-hpc/slurm-wlm-23.02` to install Slurm. Instead Slurm is pulled directly from the Ubuntu Archive. This enables us to use Slurm 23.11. Newer versions of Slurm should be published to the experimental PPA `ppa:ubuntu-hpc/experimental` rather than create a separate PPA for the Slurm packages.
2. Speeds up charm builds by using the `charm-binary-python-paclages` directive in _charmcraft.yaml_. This pulls in pre-built versions of Rust-based packages such as `cryptography` and `pydantic` rather than compile them from source with `rustc`. This should drastically reduce the amount of time it takes to run integration tests.
3. Deactivates `mysql-router` in the integration test suite. The `mysql-router` subordinate charm does not yet support Noble, so instead we integrate directly with the `mysql` charm in the integration tests. `mysql-router` supporting Noble will likely not come until after the 25.04 cycle.
4. _charmcraft.yaml_ now uses the new `base`, `platforms` syntax that was introduced with charmcraft v3 for charms that support `ubuntu@24.04` and greater.

### DIFF
--- a/charms/slurmctld/charmcraft.yaml
+++ b/charms/slurmctld/charmcraft.yaml
@@ -44,28 +44,15 @@ assumes:
   - juju
 
 type: charm
-bases:
-  - build-on:
-      - name: ubuntu
-        channel: "22.04"
-    run-on:
-      - name: ubuntu
-        channel: "22.04"
-        architectures: [amd64]
+base: ubuntu@24.04
+platforms:
+  amd64:
 
 parts:
   charm:
-    charm-requirements: ["requirements.txt"]
-    override-build: |
-      cp /usr/bin/rustc-1.80 /usr/bin/rustc
-      craftctl default
-    build-packages:
-      - libffi-dev
-      - libssl-dev
-      - rustc-1.80
-      - cargo
-      - pkg-config
-      - git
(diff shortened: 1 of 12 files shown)


dsloanm (approved):
Looks good - no problems deploying and running a test job.

I assume that stuck ubuntu@22.04 CI check was fixed by [7516ce3](https://github.com/charmed-hpc/slurm-charms/pull/51/commits/7516ce3116474b3b0fedbd382e1e3cbd2dee7fec) and can be ignored.

NucciTheBoss:
Hmm... re-running doesn't seem to refresh the state, so I might need to manipulate the branch protection rules for a hot second.

jamesbeedy:
+1 Nice!!!!

Issue #52, opened 2024-12-08
slurmdbd has an invalid PID file location configuration in `slurmdbd.conf`


NucciTheBoss:
### Bug Description

The `slurmdbd` service is configured to place its PID file in _/var/run/slurmdbd.pid_, however, the service fails to create the PID file with permission denied as the `slurm` user cannot write to the _/var/run_ directory. The `PidFile` location should be changed to `/var/run/slurmdbd/slurmdbd.pid` as _/var/run/slurmdbd_ is owned by the `slurm` user.
 
https://github.com/charmed-hpc/slurm-charms/blob/274fc856bee82c7d384e7e127636b215577d363d/charms/slurmdbd/src/constants.py#L13

### To Reproduce

1. `tox run -e integration -- --keep-models`
2. `juju switch <model-name>`
3. _Wait for all units to reach active status..._
4. `juju ssh slurmdbd/leader`
5. `systemctl status slurmdbd`
6. _See permission denied error about the PID file..._

### Environment

Juju v3.5.3 on local LXD cloud running on Ubuntu 24.04. Using the latest development version of the slurmdbd charm.

### Relevant log output

```shell
Dec 06 21:28:14 juju-9ba62c-2 systemd[1]: Started slurmdbd.service - Slurm DBD accounting daemon.
░░ Subject: A start job for unit slurmdbd.service has finished successfully
░░ Defined-By: systemd
░░ Support: http://www.ubuntu.com/support
░░
░░ A start job for unit slurmdbd.service has finished successfully.
░░
░░ The job identifier is 2812.
Dec 06 21:28:14 juju-9ba62c-2 (slurmdbd)[5821]: slurmdbd.service: Referenced but unset environment variable evaluates to an empty string: SLURMDBD_OPTIONS
Dec 06 21:28:14 juju-9ba62c-2 slurmdbd[5821]: slurmdbd: error: Unable to open pidfile `/var/run/slurmdbd.pid': Permission denied
```

### Additional context

_/var/run/slurmdbd_ is created as part of installing the `slurmdbd` Debian package from the Ubuntu Archive.

Issue #56, opened 2024-12-19
Removing new `slurmd` unit causes `hook failed: "slurmd-relation-departed"` in `slurmctld` due to invalid `scontrol update`


dsloanm:
### Bug Description

In a cluster with `slurmctld` and `slurmd` charms deployed and integrated, adding and removing a `slurmd` unit causes `slurmctld` to error with `hook failed: "slurmd-relation-departed"`.

Logs indicate the failure is due to an attempt to resume the departed node. Specifically, in `_on_write_slurm_conf()`, `self._resume_nodes(transitioning_nodes)` is called, which attempts command `['scontrol', 'update', 'nodename=juju-8e1c9b-2', 'state=resume']`, where `juju-8e1c9b-2` is the departed node. This results in a `slurm_update error: Invalid node name specified`, causing the hook failure.

The hook failure does not occur if `juju run node-configured` is run on the unit prior to removal. In that case, the unit is successfully removed without issue.

### To Reproduce

```shell
$ juju add-model charmed-hpc
Added 'charmed-hpc' model on localhost/localhost with credential 'localhost' for user 'admin'
$ juju deploy slurmctld --channel latest/edge --base ubuntu@24.04
Deployed "slurmctld" from charm-hub charm "slurmctld", revision 83 in channel latest/edge on ubuntu@24.04/stable
$ juju deploy slurmd --channel latest/edge --base ubuntu@24.04
Deployed "slurmd" from charm-hub charm "slurmd", revision 104 in channel latest/edge on ubuntu@24.04/stable
$ juju add-unit slurmd
$ juju integrate slurmctld:slurmd slurmd:slurmctld

# Wait for deployments to finish

$ juju remove-unit slurmd/1 --no-prompt
will remove unit slurmd/1
$ juju status
Model        Controller      Cloud/Region         Version  SLA          Timestamp
charmed-hpc  lxd-controller  localhost/localhost  3.5.4    unsupported  18:14:04Z

App        Version          Status  Scale  Charm      Channel      Rev  Exposed  Message
slurmctld  23.11.4-1.2u...  error       1  slurmctld  latest/edge   83  no       hook failed: "slurmd-relation-departed"
slurmd     23.11.4-1.2u...  active      1  slurmd     latest/edge  104  no       

Unit          Workload  Agent  Machine  Public address  Ports  Message
slurmctld/0*  error     idle   0        10.210.110.59          hook failed: "slurmd-relation-departed"
[...]

Issue #18, opened 2024-08-12
slurm charms should support the `sstat` command


jamesbeedy:
### Bug Description

The slurm charms currently don't support the use of the `sstat` command. Per [slurm's sstat command documentation](https://slurm.schedmd.com/sstat.html), an `acct_gather` plugin must be configured. 
<img width="1045" alt="Screenshot 2024-08-12 at 3 20 27 PM" src="https://github.com/user-attachments/assets/6a2381e4-d26b-4759-b865-382b7835612f">

### To Reproduce

na

### Environment

na

### Relevant log output

```shell
na
```

### Additional context

The slurmctld charm could support using the `sstat` command by exporting a filesystem share to the slurmd application and enabling the [acct_gather_profile/HDF5](https://slurm.schedmd.com/acct_gather.conf.html#SECTION_acct_gather_profile/HDF5)

PR #70, opened 2025-01-27
feat(slurmctld): Support InfluxDB

### LINKED ISSUE #74: Slurmdbd false positive active status
### Bug Description

Slurmdbd shows active in juju status even if it doesn't start.

Image

### To Reproduce

Deploy the slurmctld, slurmdbd, and mysql charms from edge and relate slurmdbd and mysql, but wait to set the slurmctld:slurmdbd relation until after the MySQL relation has settled.

### Environment

na

### Relevant log output

```shell
https://paste.ubuntu.com/p/kxGx4BVKXQ/
```

### Additional context

_No response_

jamesbeedy:
These changes add and modify code pertaining to the support of InfluxDB for a job profiling database.

Added pip requirements: netifaces-plus, influxdb

Changes:
* cluster_name changed to a public property of the charm and is now stored in the peer-relation application data.

* Removed default cluster-name charm config and replaced with generated name "charmed-hpc-XXXX" if no charm config is supplied.

* Added _on_start charm event to allow slurmctld-peer interface time to join after _on_install hook completes and before the slurm.conf is written (now in _on_start()).

* Added influxdb interface for relating to influxdb charm.

* Improve the influxdb relation by using relation-joined instead of relation-changed and fix an error with departing units stuck in the new_nodes stored state by removing them from new_nodes when a unit departs.

* Add new_node stored state reconciliation on slurmd_departed for nodes that have been removed from the relation but remain in new_nodes in stored state.

* Fix the slurmdbd joined/changed race by adding logic to the config rendering that will disallow the AltAuth/jwt_key from making it into the configuration until we actually have the key from slurmctld.

* Remove unused slurmd override template.

* Update slurmdbd.pid filepath and add slurmdbd override.

* Adjusted unit tests to account for new cluster_name changes.

* Added slurmctld-peer interface for storing cluster_name.

* Drive by fix for slurmdbd failing to start and false positive status with extra conditional in check_status, Fixes #74 and #75 

Fixes: #52, #56, #18, #56
Depends on: https://github.com/charmed-hpc/hpc-libs/pull/66, https://github.com/charmed-hpc/hpc-libs/pull/65, https://github.com/charmed-hpc/hpc-libs/pull/64, https://github.com/charmed-hpc/hpc-libs/pull/72

TODO: Integration tests


charms/slurmctld/src/interface_slurmctld_peer.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2025 Omnivector, LLC
+# See LICENSE file for licensing details.
+
+"""SlurmctldPeer."""
+
+# from ops import Object, RelationCreatedEvent
```
jamesbeedy, line 6:
```suggestion
```

tests/integration/test_charm.py
```diff
@@ -180,3 +189,26 @@ async def test_slurmdbd_port_listen(ops_test: OpsTest) -> None:
...
+    slurmd_unit = ops_test.model.applications[SLURMD].units[0]
+    action = await slurmd_unit.run_action("node-configured")
+    action = await action.wait()
+
+    # Get the hostname of the compute node
+    res_from_ssh = await slurmd_unit.ssh("hostname -s")
+
+    logger.info("Running a simple job to get the hostname of the compute node.")
+    sackd_unit = ops_test.model.applications[SACKD].units[0]
+    res_from_slurm_job = await sackd_unit.ssh("srun -pslurmd hostname")
```
jamesbeedy, line 213:
```suggestion
    res_from_slurm_job = await sackd_unit.ssh("srun -pslurmd hostname -s")
```

jamesbeedy:
> @jamesbeedy we're gonna take a look at your PR this week! We have some documentation stuff that we're cooking up

Sweet! Thanks!

charms/slurmctld/src/charm.py
```diff
@@ -181,9 +203,43 @@ def _on_slurmdbd_available(self, event: SlurmdbdAvailableEvent) -> None:
         self._stored.slurmdbd_host = event.slurmdbd_host
         self._on_write_slurm_conf(event)
 
-    def _on_slurmdbd_unavailable(self, _: SlurmdbdUnavailableEvent) -> None:
+    def _on_slurmdbd_unavailable(self, event: SlurmdbdUnavailableEvent) -> None:
         self._stored.slurmdbd_host = ""
-        self._check_status()
+        self._on_write_slurm_conf(event)
+
+    def _on_influxdb_available(self, event: InfluxDBAvailableEvent):
```
jamesbeedy, line 210:
```suggestion
    def _on_influxdb_available(self, event: InfluxDBAvailableEvent) -> None:
```

NucciTheBoss:
Hey @jamesbeedy, this looks good so far. Can you just do a `git rebase origin/main` real quick? Code looks good but I'm noticing that some changes in this PR here look like they're duplicating previous commits that have been merged already.

charms/slurmctld/src/charm.py
```diff
@@ -188,9 +223,46 @@ def _on_slurmdbd_available(self, event: SlurmdbdAvailableEvent) -> None:
...
+        acct_gather_params = {
+            "ProfileInfluxDBDefault": ["ALL"],
+            "ProfileInfluxDBHost": event.influxdb_host,
+            "ProfileInfluxDBUser": event.influxdb_user,
+            "ProfileInfluxDBPass": event.influxdb_pass,
+            "ProfileInfluxDBDatabase": event.influxdb_database,
+            "ProfileInfluxDBRTPolicy": event.influxdb_policy,
+        }
+        acct_gather_params.update({"SysfsInterfaces": interfaces()})
+        logger.debug(f"## _stored.acct_gather_params: {acct_gather_params}")
```
NucciTheBoss, line 243: Do we want this information logged here? It will reveal the InfluxDB password to anyone that has access to the juju log stream.

charms/slurmctld/src/interface_influxdb.py
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python3
```
NucciTheBoss, line 1: [nit]: Could you add an Omnivector copywrite notice here just so anyone coming in will know that you contributed this interface?

charms/slurmctld/src/interface_influxdb.py
```diff
@@ -0,0 +1,257 @@
...
+"""Influxdb interface."""
+import json
```
NucciTheBoss, line 3: [nit]: I like to have newline in between the module docstring and the first important statement:

```suggestion
"""Influxdb interface."""

import json
```

charms/slurmctld/src/interface_slurmd.py
```diff
@@ -286,3 +286,16 @@ def get_all_gres_info(self) -> Dict[str, Any]:
                             gres_info[node_name] = gres
 
         return gres_info
+
+    def get_active_nodes(self) -> List:
```
NucciTheBoss, line 290: [nit]: If we're grabbing just the name of the node, we could more narrowly scope this annotation to say that this method returns a list of strings rather than just a generic list:

```suggestion
    def get_active_nodes(self) -> List[str]:
```

charms/slurmctld/src/interface_influxdb.py
```diff
@@ -0,0 +1,257 @@
...
+                            # Influxdb client
+                            client = influxdb.InfluxDBClient(
+                                host=admin_info["host"],
+                                port=int(admin_info["port"]),
+                                username=admin_info["user"],
+                                password=admin_info["password"],
+                            )
+
+                            # Influxdb slurm user password
+                            influx_slurm_password = generate_random_string(20)
```
NucciTheBoss, line 141: Is there a limitation on what characters can be used for the InfluxDB password? If not, could we possibly use `secrets` rather than our own implementation of a random string generator? A 20 character string is relatively easier to brute-force than a 32 character string.

We can use the `token_urlsafe` function from the secrets module to generate passwords for the InfluxDB database:

```pyi
>>> import secrets
>>> secrets.token_urlsafe(32)
'77lIfB5SFHH-GdywGmzSrr9Tp7g9ITYiCYhDv6XMLQE'
```

charms/slurmctld/src/interface_influxdb.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+class InfluxDBEvents(ObjectEvents):
+    """InfluxDBEvents."""
+
+    influxdb_available = EventSource(InfluxDBAvailableEvent)
+    influxdb_unavailable = EventSource(InfluxDBUnavailableEvent)
+
+
+class InfluxDB(Object):
```
NucciTheBoss, line 72: [question]: What's the difference between the application-level InfuxDB relation databag and the unit-level databag? What does the application leader provide that the units don't?

charms/sackd/src/charm.py
```diff
@@ -99,13 +102,15 @@ def _on_slurmctld_available(self, event: SlurmctldAvailableEvent) -> None:
...
-            self._sackd.service.restart()
-        else:
-            self._sackd.service.start()
+        try:
+            if self._sackd.service.active():
+                self._sackd.service.restart()
+            else:
+                self._sackd.service.start()
+        except SlurmOpsError as e:
+            logger.debug(e)
```
NucciTheBoss, line 111: [nit]: This should be an error level message:

```suggestion
            logger.error(e)
```

charms/slurmctld/src/charm.py
```diff
@@ -233,6 +305,21 @@ def _on_slurmd_departed(self, event: SlurmdDepartedEvent) -> None:
...
+        active_nodes = self._slurmd.get_active_nodes()
+        logger.debug(f"Active nodes: {active_nodes}")
+
+        for node in new_nodes:
+            if node not in active_nodes:
+                logger.debug(f"Removing node: {node}")
+                new_nodes.remove(node)
+
+        # Set the remaining new nodes
+        self.new_nodes = new_nodes
```
NucciTheBoss, line 321: [question]: Remind me again what prompted adding this bit? Is it due to inconsistent state between what's reported by the relation databags and what's cached by the `new_nodes` property.

NucciTheBoss (changes requested):
Hey hey @jamesbeedy - overall I'm good with the InfluxDB interface! I left several comments but here's a TL;DR:

1. For the cluster name, I'm unsure why it needs to be tracked by a peer-relation. I figured we could use a similar approach in the slurmdbd charm where we disallow multiple units.
2. For generating a cluster name when none is provided by the cluster admin, I suggest we use the `petname` module rather than generating a random postfix.
3. Do we want to run `_on_write_slurm_conf()` at the end of the `_on_start()` hook?
4. Should we be logging the InfluxDB password?
5. Some comments around the InfluxDB interface such as using the `secrets.token_urlsafe` function for generating the InfluxDB password.

Also, could you send me the InfluxDB charm? I want to check it out :smiley: 

charms/slurmctld/src/interface_slurmctld_peer.py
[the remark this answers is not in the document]
jamesbeedy: I used the peer-relation so that if/when the slurmctld unit somehow goes away and needs to be replaced with another unit, the next unit of slurmctld will have access to the cluster-name that was initially generated by slurmctld/0.

charms/slurmrestd/src/charm.py
```diff
@@ -76,7 +76,6 @@ def _on_slurmctld_available(self, event: SlurmctldAvailableEvent) -> None:
         if (event.munge_key is not None) and (event.slurm_conf is not None):
             self._slurmrestd.munge.key.set(event.munge_key)
             self._slurmrestd.config.dump(SlurmConfig.from_str(event.slurm_conf))
-            self._slurmrestd.munge.service.restart()
```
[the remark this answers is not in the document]
jamesbeedy: ahh - nice catch!

charms/slurmctld/src/interface_slurmd.py
jamesbeedy: We need to account for the case where the empty list is returned.

charms/slurmctld/src/charm.py
jamesbeedy: https://github.com/charmed-hpc/slurm-charms/issues/56

charms/slurmctld/src/charm.py
```diff
@@ -127,7 +144,25 @@ def _on_install(self, event: InstallEvent) -> None:
...
-        self._on_write_slurm_conf(event)
+    def _on_start(self, event: StartEvent) -> None:
+        """Set cluster_name and write slurm.conf."""
+        if self.unit.is_leader() and self._slurmctld_peer.cluster_name is None:
+            if (charm_config_cluster_name := str(self.config.get("cluster-name", ""))) != "":
+                cluster_name = charm_config_cluster_name
+            else:
+                cluster_name = f"{CLUSTER_NAME_PREFIX}-{generate_random_string(4)}"
+
+            logger.debug(f"## Cluster Name: {cluster_name}")
```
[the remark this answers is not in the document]
jamesbeedy: We only need to generate a random 4 characters... I didn't feel that it warranted pulling in an additional dependency.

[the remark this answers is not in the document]
jamesbeedy: Having the "charmed-hpc" prefix on the cluster name is nice because it makes it somewhat identifiable and linkable to the charmed-hpc cluster. `'humble-quagga'` doesn't provide any context and we need to bring along another dependency.... idk

charms/slurmctld/src/charm.py
jamesbeedy: When a node goes away, it is currently still accounted for in new_nodes. We need to make sure that new_nodes only contain nodes that still exist.

charms/slurmctld/src/charm.py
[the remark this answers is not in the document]
wolsen: Is the model name a sufficient name for the fallback cluster name? It can be deployed with a name that provides context for what it actually is (up to the deployer). This also has the property that adding another unit would retain the same name (for the failure scenario mentioned below).

charms/slurmctld/src/interface_influxdb.py
jamesbeedy: Using the application databag gives us the persistence of the admin_info in the case slurmctld needs to go away and be replaced by a new unit.

charms/slurmctld/src/charm.py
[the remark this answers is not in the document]
jamesbeedy: Hey @wolsen! I do like that idea.

The operation of the entire cluster depends on this cluster-name (the tables are prefixed with the cluster-name in the database), so I was trying to create a situation where once it was set it would be almost impossible to unset or change. If we don't cache it in the peer-relation, the user could effectively set the `cluster-name` charm config and end up with a borked system. I felt that caching it the first time around and using that cached value reduced the surface area for potential misconfigurations or corner cases where the charm config gets changed. I'm +1 with using a postfix of the model name, but I still feel that caching the cluster-name in the peer-relation makes sense.

Thoughts?

charms/slurmctld/src/interface_slurmd.py
NucciTheBoss: I see what you're saying, but per [PEP 484](https://peps.python.org/pep-0484/#the-typing-module), all containers with a uniform element type can be empty. This means that `list[str]` is the correct annotation, even if `get_active_nodes(...)` can possibly return an empty list. 

`pyright` will happily let through an empty list even if the annotation is `list[str]`. If we were intentionally return an empty list every time, then we would want to use the annotation `list[Never]`:

Code sample in [pyright playground](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoByApgG5EgBQVAJkcFGkTAPoCGAxjEmcymLQGcAFAEooAWgB8UADZIBMANoKQAXQBcFKNqggmAVxAooi1VQq16rGQHdWcAcyLZ4zOQtETp7pcTJrNHV0DIxNVIA)

```python
from typing import Never


def get_active_nodes() -> list[str]:
    return []


def always_empty_list() -> list[Never]:
    return []
```

If we wanted to differentiate between a nullable and non-nullable case, we could use the annotation `list[str] | None`:

Code sample in [pyright playground](https://pyright-play.net/?code=CYUwZgBA5iAuD6BDAxrAlgNxPAdge1AGcAKQkAGzAEoIBaAPgnLUNgG1WAnAXQgB8IAOTw4QALgBQEaRHxEIAXghtuUmZzgBXTjlkEQhCGkjkQOYnIM1GABggUyQkSCA)

```python
def get_active_nodes(self) -> list[str] | None:
    nodes = []
    return nodes if len(nodes) > 0 else None
```


A bit wordy, but the TL;DR here is that `list[str]` is the correct annotation even if an empty list can be returned, and we should differentiate between `list[str]` and `None` if returning an empty list will cause the charm to break. I think we should use `list[str]` since it limits the types that can be added to the returned list in `get_active_nodes(...)`, and we get autocompletion benefits in our IDEs.

charms/slurmctld/src/charm.py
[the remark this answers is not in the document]
jamesbeedy: I'm agree with the sentiment, but it doesn't match real life deployment use cases. For example, I've grown accustomed to deploying all slurm admin services to a model called `slurm-admin`.

charms/slurmctld/src/utils.py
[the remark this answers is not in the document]
jamesbeedy: I believe that adding another library to the project is unnecessary when we can generate a few characters using a built-in library.

charms/slurmctld/src/charm.py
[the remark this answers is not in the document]
NucciTheBoss: Typically we deploy the controller in the same model as the compute nodes which is where I got the idea from, however, you do bring up a valid point in that we can't really control how people downstream will consume the charms. Some will want the control plane in another model while others may co-locate all the charms together in one model.

I have some ideas about generating the unique name that don't involve the model or the custom generator function.

charms/slurmctld/src/charm.py
```diff
@@ -127,7 +144,42 @@ def _on_install(self, event: InstallEvent) -> None:
...
-        self._on_write_slurm_conf(event)
+    def _on_start(self, event: StartEvent) -> None:
+        """Set cluster_name and write slurm.conf.
+
+        Note: The start hook can execute multiple times in a charms lifecycle,
+              for example, after a reboot of the underlying instance. This code safeguards
+              against the potentiality of changing the cluster_name in subsequent start hook
+              executions by applying logic that ensures the cluster_name is only set on the
+              first execution of this hook, we log and return on any subsequent start hook
+              event executions.
```
NucciTheBoss, line 155: [nit]: just a docstring format request:

```suggestion
        Notes: 
            - The start hook can execute multiple times in a charms lifecycle,
              for example, after a reboot of the underlying instance. This code safeguards
              against the potentiality of changing the cluster_name in subsequent start hook
              executions by applying logic that ensures the cluster_name is only set on the
              first execution of this hook, we log and return on any subsequent start hook
              event executions.
```

charms/slurmctld/src/interface_slurmctld_peer.py
```diff
@@ -0,0 +1,63 @@
...
+            logger.debug(
+                "## `slurmctld-peer` relation not available yet, cannot get cluster_name."
+            )
+        return cluster_name
+
+    @cluster_name.setter
+    def cluster_name(self, name: str) -> None:
+        """Set the cluster_name on app relation data."""
+        if not self.framework.model.unit.is_leader():
+            raise SlurmctldPeerError("Only the leader can set the cluster_name.")
```
NucciTheBoss, line 56: Do we need to raise an exception here? We could just log that non-leader units can't set the cluster name and `return`?

```suggestion
        if not self.framework.model.unit.is_leader():
            logger.debug("only leader can set the Slurm cluster name")
            return
```

This way we don't need to worry about catching an exception, and invocations of this setter don't always need to be nested within an `is_leader()` block.

charms/slurmctld/src/charm.py
```diff
@@ -127,7 +144,42 @@ def _on_install(self, event: InstallEvent) -> None:
...
+        """
+        if self.unit.is_leader():
+            if self._slurmctld_peer.cluster_name is None:
+                if (charm_config_cluster_name := str(self.config.get("cluster-name", ""))) != "":
+                    cluster_name = charm_config_cluster_name
+                else:
+                    cluster_name = f"{CLUSTER_NAME_PREFIX}-{generate_random_string(4)}"
```
NucciTheBoss, line 162: Could we just replace `generate_random_string(...)` with `token_urlsafe`? That way we don't need an external dependency like `petname` or have to incur any technical debt from carrying our own custom module:

```suggestion
                    cluster_name = f"{CLUSTER_NAME_PREFIX}-{secrets.token_urlsafe(3)}"
```

_`token_urlsafe(3)` generates 4 characters:_

```pyi
>>> CLUSTER_NAME_PREFIX = "charmed-hpc"
>>> f"{CLUSTER_NAME_PREFIX}-{secrets.token_urlsafe(3)}"
'charmed-hpc-Nt9c'
```

charms/slurmctld/requirements.txt
```diff
@@ -1,2 +1,4 @@
 ops==2.17.1
 slurmutils<1.0.0,>=0.11.0
+influxdb==5.3.2
+netifaces-plus==0.12.4
```
NucciTheBoss, line 4: [foresight]: eventually we should think about replacing `netifaces` given that the original was deprecated and this fork is maintained by one person seemingly.

`netplan` ended up just reimplementing what they specifically needed in their own _utils.py_ module, but we can solve this later: https://github.com/canonical/netplan/pull/503

NucciTheBoss (changes requested):
Just a few small suggestions and then I'm ready to merge this PR.

Let me know if you have any questions/comments!

jamesbeedy:
@NucciTheBoss all good?

NucciTheBoss (approved):
LGTM! :shipit: 

Thank you for putting in the work here @jamesbeedy :rocket:

PR #73, opened 2025-01-28
fix(ci): pin charmcraft to version 3.2.3

NucciTheBoss:
This should fix the CI issues we've been seeing since charmcraft 3.3.2 was released to the `latest/stable` channel. I just modified the CI and Release pipelines to just install and hold the version of charmcraft to 3.2.3. We'll still need to deep dive on what the longer term solution is, but at least we will be able to get the latest batch of PRs through.

Bug opened on the upstream `charmcraft` repository:

* https://github.com/canonical/charmcraft/issues/2125

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-# Copyright 2024 Canonical Ltd.
+# Copyright 2024-2025 Canonical Ltd.
 #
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Looks good!

NucciTheBoss:
Okay, looks like this should hopefully fix our current CI issues. @jamesbeedy @dsloanm you should be able to sync into your PRs after merging.

PR #81, opened 2025-02-11
docs: add SECURITY.md

jedel1043:
Adds a simple SECURITY.md about our security policy for the repository.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,33 @@
+# Security policy
+
+## Supported versions
+
+Security updates for the Slurm charms will be released after vulnerabilities have been identified and fixed to
+the `latest/edge` and `latest/stable` release channels.
+
+## Reporting a vulnerability
+
+Please provide a description of the issue, the steps you took to create the issue, affected versions, and, if known,
+mitigations for the issue.
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,33 @@
+# Security policy
+
+## Supported versions
+
+Security updates will be released after vulnerabilities have been identified and fixed. At this point in time,
+we will only release security patches to the `latest/edge` and `latest/candidate` charm release channels.
```
NucciTheBoss, line 6: Same as with `filesystem-charms`, IIRC we don't currently use the `latest/candidate` channel:

```suggestion
Security updates for the Slurm charms will be released after vulnerabilities have been identified and fixed to the `latest/edge` and `latest/stable` release channels.
```

We'll eventually want to sit down after mid-cycle and identify our release strategy. E.g. what version do we track, how will we split charms into channels, etc. However, what we have now is good!

NucciTheBoss (changes requested):
Looks great! Just a suggested revision for supported release channels.

NucciTheBoss (approved):
LGTM! :shipit:

PR #90, opened 2025-03-24
chore: migrate charms to `hpc-libs` charm library

jedel1043:
# Pre-submission checklist

## Summary of changes

Migrates all our charms from using the charm libraries from `hpc-libs` to use the `hpc-libs` python package.

## Docs

Change is entirely internal, so it doesn't need documentation changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,13 +21,14 @@ dependencies = [
     "influxdb==5.3.2",
     "netifaces-plus==0.12.4",
     "nvidia-ml-py==12.560.30",
+    "hpc-libs",
 ]
 
 [project.optional-dependencies]
 dev = [
     # Tests
     "ops[testing]~=2.19",
-    "cryptography~=43.0.1",
+    "cryptography~=44.0.1",
     "distro==1.9.0",
     "python-dotenv~=1.0.1",
     "pycryptodome==3.20.0",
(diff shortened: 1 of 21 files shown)


pyproject.toml
```diff
@@ -63,14 +64,13 @@ members = ["charms/*"]
 
 [tool.uv.sources]
 ubuntu-drivers-common = { git = "https://github.com/canonical/ubuntu-drivers-common", rev = "554b91edfd3699625dbed90f679abb31a897b76e" }
+hpc-libs = { git = "https://github.com/charmed-hpc/hpc-libs", rev = "1c1fd91bcc7de71fd9da7bc077f88757c81d14ba" }
```
NucciTheBoss, line 67: [question]: do we need the git hash here because we don't publish tagged revisions for `hpc_libs`?

Out of scope for this PR, but we should discuss the overall structure of `hpc_libs` going forward. We'll likely want to expose exports in the top-level *\_\_init\_\_.py* file, and eventually take `is_container` out and move it to a different module since `systemd-detect-virt` is a relatively generic utility that we can use across multiple projects.

NucciTheBoss (commented):
Just one question and then I'm ready to merge this PR :smiley:

pyproject.toml
jedel1043: > do we need the git hash here because we don't publish tagged revisions for hpc_libs

Yep, pretty much.

NucciTheBoss (approved):
LGTM! :shipit:  

We'll figure out how we can make `hpc-libs` a local dependency in the future, and decide on the structure of the library, but overall this looks great!

PR #95, opened 2025-04-03
feat: automatically manage binary packages from repository.py

jedel1043:
# Pre-submission checklist

## Summary of changes

Adds functionality on `repository.py` to inject the required binary packages onto the `charmcraft.yaml` files instead of having to manually manage them.

## Docs

Internal tooling improvement, which doesn't require documentation.

### DIFF
--- a/repository.py
+++ b/repository.py
@@ -18,7 +18,7 @@ import itertools
 from pathlib import Path
 from threading import Thread
 from dataclasses import dataclass
-from collections.abc import Iterable
+from collections.abc import Iterable, Mapping
 from typing import Any
 
 import yaml
@@ -28,6 +28,7 @@ BUILD_PATH = ROOT_DIR / "_build"
 CHARMS_PATH = ROOT_DIR / "charms"
 PYPROJECT_FILE = "pyproject.toml"
 CHARMCRAFT_FILE = "charmcraft.yaml"
+LOCK_FILE = "uv.lock"
 LIBS_CHARM_PATH = BUILD_PATH / "libs"
 
 
@@ -150,6 +151,7 @@ class Repository:
 
     def __init__(self) -> "Repository":
         """Load the monorepo information."""
+        UV.run_command(["lock", "--quiet"])
         try:
             with (ROOT_DIR / PYPROJECT_FILE).open(mode="rb") as f:
                 project = tomllib.load(f)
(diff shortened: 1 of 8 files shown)


repository.py
```diff
@@ -262,8 +305,12 @@ def stage_charm(
     if not dry_run:
         remove_dir_if_exists(charm.build_path)
         shutil.copytree(charm.path, charm.build_path, dirs_exist_ok=True)
-        if charm.metadata.get("charm-libs"):
-            CHARMCRAFT.run_command(["fetch-libs"], cwd=charm.build_path)
+        try:
+            with open(charm.build_path / CHARMCRAFT_FILE, "wt") as f:
+                yaml.safe_dump(charm.metadata, f, sort_keys=False)
+
```
NucciTheBoss, line 311: [nit]: I generally try to keep `try/except` statements together visually:

```suggestion
```

repository.py
```diff
@@ -194,7 +217,11 @@ def __init__(self) -> "Repository":
 
 
 def load_charm(
-    charm: Path, external_libraries: [CharmLibrary], internal_libraries: [CharmLibrary]
+    charm: Path,
+    external_libraries: [CharmLibrary],
+    internal_libraries: [CharmLibrary],
+    binary_packages: dict[str, str],
+    uv_lock: dict[str, Any],
```
NucciTheBoss, line 224: [nit]: generally it's preferred to use `Mapping` for annotation function arguments rather than `dict` explicitly. It makes it easier to pass objects that inherit from `Mapping` to functions. [Here's relevant bit of Python documentation that I'm referencing](https://docs.python.org/3/library/typing.html#typing.Dict):

```suggestion
    binary_packages: Mapping[str, str],
    uv_lock: Mapping[str, Any],
```

_Note that you'll need to import `Mapping` from `collections.abc`_.

repository.py
```diff
@@ -208,6 +235,22 @@ def load_charm(
...
+    pending = [charm.name]
+    while pending:
+        package = pending.pop()
+        deps.add(package)
+        for pkg_dep in next(
+            (pkg.get("dependencies", []) for pkg in uv_lock["package"] if pkg["name"] == package)
+        ):
+            if pkg_dep["name"] not in deps:
+                deps.add(pkg_dep["name"])
+                pending.append(pkg_dep["name"])
```
NucciTheBoss, line 248: Could you add a comment here or docstring to the `load_charm` function that explains why we need this recursive while loop? I know it's because `uv` doesn't resolve the dependencies of dependencies, but I could totally see someone without our context coming in and being like "wahhh :thinking:" 

We'll be thanking ourselves in six months if we need to rework this code at one point and we're trying to remember we're running this loop to analyze uv.lock files.

repository.py
```diff
@@ -147,9 +148,11 @@ class Repository:
     charms: [Charm]
     external_libraries: [CharmLibrary]
     internal_libraries: [CharmLibrary]
+    binary_packages: dict[str, str]
```
NucciTheBoss, line 151: [nit]: I _think_ we want to use `Mapping` here. Preference isn't as clear for dataclasses as compared to arguments:

```suggestion
    binary_packages: Mapping[str, str]
```

NucciTheBoss (changes requested):
This will be great for improving our dev process! Will make it much easier to manage Rust dependencies without having to manually update every single _charmcraft.yaml_.

Just a few suggestions around type annotations and comments for additional context about why we're doing certain things in the `load_charm` function.

repository.py
jedel1043: Deleted the field instead because I realized we don't need it anymore.

PR #100, opened 2025-05-06
chore: bump juju terraform provider version

jedel1043:
# Pre-submission checklist

## Summary of changes

Bumps the Juju terraform provider version to 0.19.0.

## Docs

Just an internal change.

### DIFF
--- a/charms/sackd/terraform/versions.tf
+++ b/charms/sackd/terraform/versions.tf
@@ -16,7 +16,7 @@ terraform {
   required_providers {
     juju = {
       source  = "juju/juju"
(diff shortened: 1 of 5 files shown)


NucciTheBoss (approved):
One could argue that the PR is a bit pedantic since I would imagine that `tofu init` would default to using the provider version specified in the subordinate charm Terraform modules since the `>= 0.13.0` version constraint is compliant with the `>= 0.19.0`.

However, I do think it's good for us to state that we expect our users to be using the latest version of the Juju Terraform provider since there's no API compatibility guarantees right now.

PR #130, opened 2025-07-22
feat(all): migrate to slurmutils v1 and local `slurm_ops` subpackage

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

This PR implements the local `slurm_ops` subpackage so that we no longer need to pull `slurm_ops` from `hpc_libs`, and it migrates the Slurm charms to using the slurmutils v1 API. It also lays the initial groundwork for the required peer interface changes that will be used by the slurmctld charm for configuring HA.

__Key things to note:__

- The existing integration interfaces have been modified to use `slurmutils` v1. They will eventually be replaced by the interfaces in `hpc-libs[interfaces]`
- Since this PR is the first step on our journey to implement HA and dynamic nodes within the Slurm charms, it strips out several features:
  - The `slurmd` nodes no longer share Gres device data over the integration with `slurmctld`.
  - The `slurmctld` charm no longer manages the _gres.conf_ file. _gres.conf_ management is unnecessary when using dynamic compute nodes as the compute nodes will advertise the Gres devices on the machine.
- This PR removes all `Harness`-based unit tests. I added some `Scenario`-based tests for the `slurmctld-peer` integration interface, but my plan is to gradually re-add these unit tests back as `Scenario`-based tests while we merge in HA support and the dynamic nodes work. There will be several changes to how some event handlers work, so I removed the tests so that it will make it easier to write new tests that reflect the new implementation.

This PR is related to our ongoing work to merge HA and dynamic nodes into the Slurm charms.

## Docs

Or:

This PR is against the `experimental` branch, so it's not meant to be used for anything other than development of major features.


charms/slurmctld/src/charm.py
```diff
@@ -153,37 +154,31 @@ def _on_install(self, event: InstallEvent) -> None:
...
-
-                logger.debug(f"Cluster Name: {cluster_name}")
-
-                try:
-                    self._slurmctld_peer.cluster_name = cluster_name
-                except SlurmctldPeerError as e:
-                    self.unit.status = BlockedStatus(e.message)
-                    logger.error(e.message)
-                    event.defer()
-                    return
```
dsloanm, line 181: Yep, looks good. This is a nice incremental step towards bringing the HA work in.

charms/slurmctld/src/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+        """Set the controller peer data in the `slurmctld-peer` application databag.
+
+        Args:
+            data: Controller peer data to set in the peer integration's application databag.
+
+        Warnings:
+            - Only the `slurmctld` application leader can set controller peer app data.
+        """
+        integration = self.get_integration()
+        if not integration:
```
dsloanm, line 87: I think we always expect `get_integration()` to find an integration so it would make sense for it to raise an exception if it doesn't, rather than return `None`. Then we could have something like:

```python
try:
    integration = self.get_integration()
    integration.save(data, self.app)
    _logger.info("set data")
except:
    _logger.info("did not set data")
```

Forgiveness rather than permission to do `integration.save()`, and similar for the `integration.load()` below. This would require an `hpc-libs` change but I think we should consider it.

charms/slurmctld/src/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+        """Get the unique cluster name from the application integration data."""
+        if data := self.get_controller_peer_app_data():
+            return data.cluster_name
+        else:
+            return None
+
+    @cluster_name.setter
+    @leader
+    def cluster_name(self, value: str) -> None:
+        self.set_controller_peer_app_data(ControllerPeerAppData(cluster_name=value))
```
dsloanm, line 128: Much cleaner than having leader checks and `json.dumps` in the method body!

Thinking a bit ahead, does setting a new `ControllerPeerAppData` here create any difficulties for adding the `StateSaveLocation` (and JWT key path) to the relation as part of the HA work? The HA PR currently sets a default path in the relation app data (`/var/lib/slurm/checkpoint`) then updates it (`/mnt/slurmctld-statefs/checkpoint`) when the `filesystem-client` is integrated and data is migrated to the shared storage. If we create a new `ControllerPeerAppData` on each setter call, how do we update just part of the relation data without changing the rest?

pkgs/slurm-ops/.python-version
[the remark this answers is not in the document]
NucciTheBoss: The _.python-version_ file was added when I ran the `uv init --lib pkgs/pkg-name` command in the contributing guide. I don't _think_ it's required, but if it isn't and we don't want it declared in the subpackage, we should add a note in the CONTRIBUTING.md file that mentions that this dotfile should be removed. 

Fwiw I also noticed that I accidentally firebombed that section when we updated to the new CONTRIBUTING.md format, so expect a PR against `main` soon to add that back lol :cold_face:

charms/slurmctld/src/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+@dataclass(frozen=True)
+class ControllerPeerAppData:
+    """Data provided by the primary/leader Slurm controller, `slurmctld`.
+
+    Attributes:
+        cluster_name: The unique name of this cluster.
+
+    Warnings:
+        - `cluster_name` can only be set when the peer integration is created. The cluster name
+           becomes permanent after the `SlurmctldPeerConnectedEvent` is handled.
```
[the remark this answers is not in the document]
NucciTheBoss: We can revise it to say something along the lines of "warning - it's bad to change you cluster name after it's been set as it will corrupt Slurm's state save data, so you should only set it once it an event handler such as when the slurmctld peer relation is first created."

charms/slurmctld/src/integrations.py
NucciTheBoss: I'd consider that out of scope for this PR since this PR is primarily focused on introducing the local `slurm_ops` changes, but I needed the initial peer interface updates so that the integration tests would work. 

That said, the reason why I prefer the `None` check is because typically comparing against `None` is faster than exception handling:

```python
import logging
import os
import sys
import timeit

logger = logging.getLogger()

fout = open(os.devnull, "w")
sys.stderr = fout


def lookup_check_if_none():
    data = {1: "some cool integration"}
    integration = data.get(2)
    if integration is None:
        logger.error("no integration found :(")


def lookup_handle_key_error():
    data = {1: "some cool integration"}
    try:
        _ = data[2]
    except KeyError:
        logger.error("no integration found :(")


if __name__ == "__main__":
    print(
        f"`lookup_check_if_none` ttc: {timeit.timeit(lambda: lookup_check_if_none(), number=10000)}"
    )
    print(
        f"`lookup_handle_key_error` ttc: {timeit.timeit(lambda: lookup_handle_key_error(), number=10000)}"
    )

# Output:
# `lookup_check_if_none` ttc: 0.0709606949967565
# `lookup_handle_key_error` ttc: 0.07322793199273292
```

Also, you can call `get_integration` whenever in the charm, we just only call it when there's a high likelihood that the integration does exist.

NucciTheBoss: The `get_integration(...)` method is effectively a forward for `self.charm.model.get_relation(...)` where the integration name is already set to the integration name the interface class was initialized with. `get_relation(...)` doesn't return an error if the integration is not found, so we'd need to decide if we want to make `get_integration` raise an exception if `get_relation(...)` returns `None`. We'd would likely need to add a custom exception like`"IntegrationNotFoundError` or something similar.

pkgs/slurm-ops/src/slurm_ops/core/config.py
```diff
@@ -0,0 +1,185 @@
...
+    BaseEditor,
+    CGroupConfigEditor,
+    GresConfigEditor,
+    OCIConfigEditor,
+    SlurmConfigEditor,
+    SlurmdbdConfigEditor,
+)
+
+
+class ConfigManager(ABC):
```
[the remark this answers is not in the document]
NucciTheBoss: Implemented the generic `SlurmConfigManager` class in https://github.com/charmed-hpc/slurm-charms/pull/130/commits/61fa8d2ad64e9cbf1639d14a7e6217a055499bae. Let me know what you think!

charms/slurmctld/src/integrations.py
NucciTheBoss: Either way, since we've identified these changes as out of scope for this PR, let's continue this conversation in a new PR which will introduce interface changes in `hpc_libs`.

NucciTheBoss:
Epic! @dsloanm can you look over this PR tomorrow to ensure that all your comments have been addressed?

dsloanm (approved):
Yep, good to go. I've created an issue on the hpc-libs repo for `get_integration()`: https://github.com/charmed-hpc/hpc-libs/issues/101

PR #177, opened 2026-02-09
feat(slurmd): implement `default-node-state` and `default-node-reason`

### LINKED ISSUE #164: Implement `default_node_state` and `default_node_reason` configuration options
This issue is to track the second half of UHPC004 since the `set-node-state` action was already implemented.

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

This PR implements the slurmd charm parts of UHPC004. It adds the `default_node_state` and `default_node_reason` configuration options to the `slurmd` charm.

This PR also refactors bits of the `slurmd` charm to start introducing our pseudo-MVC design pattern. Key changes here include:

1. Eliminating the `reconfigure` decorator from the `slurmd` charm, and instead use a `reconfigure` method defined as part of the `SlurmManager` class from `slurm-ops`.
2. Move Gres resource detection with `pynvml` to `SlurmdManager`. The GPU installation code is still within the slurmd charm, but the method for building the `Gres` resource configuration is now a method of `SlurmdManager`.
3. Move `get_node_info` to `SlurmdManager` and rename to `get_machine_info`.

Another change this is introduced is `ConfigManager` class for managing configuration of the `slurmd` charm. A pydantic dataclass is used to validate the slurmd charm application's configuration. Currently, we only look at part of the configuration since `slurmutils` data models aren't compatible with pydantic.

- Closes #164

## Docs

Or:

These new configuration options will be documented as part of a larger howto for managing Slurm with the Slurm charms. I'll be working on these docs next.

### DIFF
--- a/charms/slurmd/src/charm.py
+++ b/charms/slurmd/src/charm.py
@@ -36,6 +36,7 @@ from hpc_libs.interfaces import (
     wait_unless,
 )
 from hpc_libs.utils import StopCharm, refresh
+from pydantic import ValidationError
 from slurm_ops import SlurmdManager, SlurmOpsError
 from slurmutils import ModelError, Node
 from state import check_slurmd, reboot_if_required, slurmd_installed
@@ -53,9 +54,20 @@ class SlurmdCharm(ops.CharmBase):
 
         self.slurmd = SlurmdManager(self.app.name, snap=False)
         try:
-            self.configmgr = ConfigManager.load(self.config, self.app.name)
-        except SlurmOpsError as e:
-            self.unit.status = ops.BlockedStatus(e.message)
+            self.configmgr = self.load_config(ConfigManager, partition_name=self.app.name)
+        except ValidationError as e:
+            logger.error(e)
+            self.unit.status = ops.BlockedStatus(
+                "Configuration option(s) "
+                + ", ".join(
+                    [
+                        f"'{option.replace('_', '-')}'"  # type: ignore
+                        for error in e.errors()
+                        for option in error.get("loc", ())
+                    ]
+                )
+                + " failed validation. See `juju debug-log` for details"
+            )
             return
 
         framework.observe(self.on.install, self._on_install)
@@ -148,8 +160,7 @@ class SlurmdCharm(ops.CharmBase):
         self.slurmd.key.set(data.auth_key)
         self.slurmd.conf_server = data.controllers
 
-        # Set default state and reason if this compute node is being added to a Slurm cluster
-        # and not just updated an updated controller list or Slurm key.
+        # Set default state and reason if this compute node is being added to a Slurm cluster.
         params = {}
         if not self.slurmd.exists():
             params = {
--- a/charms/slurmd/src/config.py
+++ b/charms/slurmd/src/config.py
@@ -17,9 +17,7 @@
 import logging
 from typing import Literal
 
-from ops import ConfigData
-from pydantic import BaseModel, ConfigDict, ValidationError, ValidationInfo, field_validator
-from slurm_ops import SlurmOpsError
+from pydantic import BaseModel, ConfigDict, ValidationInfo, field_validator
 from slurmutils import ModelError, Partition
 
 _logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


charms/slurmd/src/config.py
```diff
@@ -15,151 +15,128 @@
...
+                    [f"'{option}'" for error in e.errors() for option in error.get("loc", ())]
+                )
+                + " failed validation. See `juju debug-log` for details"
+            )
+
+    @model_validator(mode="after")
+    def update_defaults(self) -> Self:
+        """Update default values for configuration options."""
+        if self.default_node_state != "idle" and not self.default_node_reason:
+            object.__setattr__(self, "default_node_reason", "n/a")
```
dsloanm: Making this class frozen and then circumventing it with `object.__setattr__` seems unnecessarily hacky. The factory pattern with `load` could also be simplified. How about making `ConfigData` a plain class with no methods:

```python
@dataclass(frozen=True)
class ConfigData:
    default_node_state: Literal["idle", "down"] = "idle"
    default_node_reason: str = ""
```

Then putting the logic here:

```python
class ConfigManager:
    """Configuration manager for the `slurmd` charmed operator."""

    def __init__(self, charm: "SlurmdCharm", /) -> None:
        self._charm = charm
        
        node_state = charm.config.get("default_node_state", "idle")
        node_reason = charm.config.get("default_node_reason", "")
        if node_state != "idle" and not node_reason:
            node_reason = "n/a"

        try:
            self._data = ConfigData(
                default_node_state=node_state,
                default_node_reason=node_reason,
            )
        except ValidationError as e:
            ...
```

charms/slurmd/src/config.py
```diff
@@ -15,151 +15,128 @@
...
+
+        Args:
+            partition: Partition configuration data to set.
+        """
+        if self._charm.slurmctld.is_joined():
+            _logger.info("updating '%s' partition configuration", self._charm.app.name)
+            _logger.debug(
+                "'%s' partition configuration:\n%s", self._charm.app.name, plog(partition.dict())
+            )
+            self._charm.slurmctld.set_compute_data(ComputeData(partition=partition))
```
dsloanm: `ConfigManager` and `SlurmdProvider` both providing methods to interact with the integration databag is potentially confusing. When writing a hook, it's not clear when `self.slurmctld.set_compute_data` should be used over a `self.configmgr` method in `charm.py` code, for example. Also, given the config values and logic are fairly minimal, I think we might be over-abstracting here.

If all interactions with the databag should be through `ConfigManager`, and they should if we do introduce a manager, then moving the construction of `self.slurmctld = SlurmdProvider` to `ConfigManager` to help clarify this could be an option. However, that would complicate `framework.observe`-ing the hooks and we'd probably find we need the observer pattern.

Alternatively, we could move the config methods to `charm.py` - i.e. `self.default_node_conf()` and `self.update_partition_config()` (if `set` and `get` are merged per another comment). This would be my preference, given the code deals with charming concerns and is fairly minimal.

NucciTheBoss:
@dsloanm PR is ready for round 2.

The biggest change in these reworked commits is how we're using `ConfigManager`. Instead of being a holder of utility methods such as `get_partition` and `set_partition`, it now only manages the slurmd charm's configuration set using `juju config`. I added additional logic so that `ConfigManager` can manage the custom partition configuration information. I used pydantic's field validator mechanism to validate the partition configuration, and if the validator fails, a `ValidatorError` is raised that will state that the provided partition configuration is bad. I still think we should use pydantic's field validator mechanism rather than intercept configuration values in the `load` class method body because then we only have to handle a single `ValidationError` if a particular configuration field fails validation.

Other noticeable changes are that I modified the `SlurmdManager` class to accept the application name in its `__init__` method. I really don't like the idea of having "dangling utility methods" on the `SlurmdCharm` class, so I modified `build_node` so that it would automatically add the required features to the node configuration definition. I also removed some redundant parameterized unit tests from `slurm-ops`. We don't really to test every possible Slurm service name we can pass to `SlurmManager` because we can infer that if the logic around one Slurm service name is correct, then the logic is correct around the other Slurm service names.

charms/slurmd/src/charm.py
```diff
@@ -242,21 +220,40 @@ def _on_set_node_config_action(self, event: ops.ActionEvent) -> None:
...
-
-        self.service_needs_restart = True
-        # TODO: Return updated `slurm.conf` once `reconfigure` is refactored.
+            self.slurmd.conf = self.slurmd.build_node()
+
+        config = self.slurmd.conf
+        config.update(custom)
+        self.slurmd.conf = config
+
+        if self.slurmd.exists():
```
jedel1043:
```suggestion
        if !self.slurmd.exists():
            event.set_results({"accepted": True})
            return
```

Never indent gang?

pkgs/slurm-ops/src/slurm_ops/slurmd.py
```diff
@@ -92,3 +108,151 @@ def user(self) -> str:
...
+
+            cmd.append(f"reason={reason}")
+        elif state == "idle" and reason:
+            _logger.warning(
+                "the 'idle' state does not require a reason to be set. "
+                f"not updating node's {self.name} 'reason' field to '{reason}'."
+            )
+
+        _logger.info("setting state of node %s to state '%s'", self.name, state)
+        scontrol(*cmd)
```
jedel1043: Could this fail if the service hasn't finished restarting before the command gets run?

Maybe this is a bit paranoic, but just wanted to raise this in case it wasn't considered yet.

NucciTheBoss: The expected behavior of this method is to raise a `SlurmOpsError` if something goes wrong, so I wouldn't get too caught up opn trying to be defensive as we want to surface errors if there is an issue here.

That said, `scontrol` won't fail because `systemctl` won't return until `ExecStart` has completed which is what starts the `slurmd` service. That's why `systemctl start slurmd` used to hang for so long in the original Slurm charms implementation. `systemctl` would wait for up to 10 minutes while our wrapper script tried to start slurmd as a child process in the background.

charms/slurmd/src/charm.py
```diff
@@ -40,40 +34,46 @@
...
+            self.unit.status = ops.BlockedStatus(
+                "Configuration option(s) "
+                + ", ".join(
+                    [
+                        f"'{option.replace('_', '-')}'"  # type: ignore
+                        for error in e.errors()
+                        for option in error.get("loc", ())
+                    ]
+                )
+                + " failed validation. See `juju debug-log` for details"
```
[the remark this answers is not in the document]
jedel1043: Kinda bad wording from my part. What I meant was if `_logger.error(e)` already pretty prints `ValidationError`, which avoids having to manually do that here.

dsloanm (approved):
Think this is good to go now

PR #202, opened 2026-04-02
feat(ci): enable support for internal and external subpackages

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

This PR implements specification [UHPC009](https://github.com/charmed-hpc/specs/blob/main/specs/UHPC%20009%20-%20Distributing%20Slurm%20interfaces%20as%20Python%20packages/uhpc009.md) by patching the _repository.py_ script to support both internal and external packages. New subpackages/"uv workspaces" can be placed under _internal_ or _pkg_ depending on whether they should be made publicly available. 

#### Other changes

It was discovered as part of patching _repository.py_ that a good chunk of the unit tests for `slurm-ops` were broken as the implementation changed overtime, but mocks had not been updated. Now that packages are included with `just repo unit` invocations, these issues will be caught going forward.

- Implementation of specification [UHPC009](https://github.com/charmed-hpc/specs/blob/main/specs/UHPC%20009%20-%20Distributing%20Slurm%20interfaces%20as%20Python%20packages/uhpc009.md)

## Docs

Or:

This is non-user facing change that modifies how we publish our shared artifacts.


repository.py
```diff
@@ -154,10 +154,13 @@ def charm_path(self) -> Path:
...
-    internal_packages: Iterable[Package]
-    external_libraries: Iterable[CharmLibrary]
-    internal_libraries: Iterable[CharmLibrary]
+    charms: Collection[Charm]
+    external_libraries: Collection[CharmLibrary]
+    internal_libraries: Collection[CharmLibrary]
+    libraries: Collection[CharmLibrary]
+    external_packages: Collection[Package]
+    internal_packages: Collection[Package]
+    packages: Collection[Package]
```
jedel1043: Wondering if we really need to have `external_libraries` + `internal_libraries` + `libraries`. Having only external + internal would avoid the problem of forgetting to update `libraries` whenever `external_libraries` and `internal_libraries` gets updated, and viceversa. Same for `packages`.

repository.py
```diff
@@ -154,10 +154,13 @@ def charm_path(self) -> Path:
...
-    charms: Iterable[Charm]
-    internal_packages: Iterable[Package]
-    external_libraries: Iterable[CharmLibrary]
-    internal_libraries: Iterable[CharmLibrary]
+    charms: Collection[Charm]
+    external_libraries: Collection[CharmLibrary]
+    internal_libraries: Collection[CharmLibrary]
+    libraries: Collection[CharmLibrary]
+    external_packages: Collection[Package]
+    internal_packages: Collection[Package]
```
jedel1043: I would intentionally avoid using "external" and "internal" nomenclature for packages, because they're not the same as "external" libraries (Charmhub libraries consumed by us) and "internal" libraries (Charmhub libraries maintained by us).

Would be better to use `public` and `private` packages in this case, to better reflect that the packages are _all_ maintained by us.

repository.py
jedel1043: This could also be two getters on `Repository` tbh.

repository.py
```diff
@@ -588,30 +615,67 @@ def main_cli():
...
+    charms: Collection[Charm],
+    packages: Collection[Package],
     repository: Repository,
     clean: bool = False,
     dry_run: bool = False,
     **kwargs,
 ):
     """Stage the specified charms into the build directory."""
+    if len(charms) == 0:
+        raise RepositoryError("Please provide charms to stage")
```
jedel1043, line 648: Isn't this case kinda impossible? Since `main` takes care of passing all available charms to this function if the list of provided charms is empty.

jedel1043: FWIW I think it would be nicer if we had a `strict_filter` method on `Repository` that gets the list of required charms from all the available charms, or errors if some provided charm name doesn't exist, but that can be postponed for a future `repository.py` cleanup.

jedel1043 (changes requested):
Looks very good! Couple of comments about the implementation, but I agree with most of the changes.

repository.py
NucciTheBoss: I was just building off of the existing nomenclature used in the script, but we can change the terminology. One could argue that "public" means that we're pulling public libraries from Charmhub and "private" means that we're using private libraries vendored inside the charm, but frankly that's overthinking it imho (also charm libraries are effectively legacy at this point).

I do like the public and private terminology better as I think "public" better reflects that a package is meant to be consumed publicly while private packages are not.

repository.py
NucciTheBoss: Hmm... aren't these values largely static? They're only computed at the beginning of the script, and they don't get modified at any point after the initial `Repository` object is created and handed off to `context`.

I really only added the `libraries` and `packages` attributes because I saw that there were multiple places where we were calling `itertools.chain(...)` to iterate over both the `internal_*` and `external_*` attributes  together, and I just thought it was simpler to compute these values when the `Repository` object is created rather than be clever with `itertools.chain()` since we already know we want the combined list of both attributes.

We could implement lazy loading with getters where the libraries/package list isn't computed until the getters are accessed, but I'm wondering if that's just a premature optimization for our current needs since `packages` and `libraries` are static after the initial `Repository` object is created in `main_cli`.

repository.py
NucciTheBoss: It is possible because of how I'm handling the detection of a target's type in `main_cli`:

```python
        # Detect if provided target is a charm or package.
        for target in targets:
            try:
                context["charms"].append(validate_charm(target, repository))
                continue
            except RepositoryError:
                pass

            try:
                context["packages"].append(validate_package(target, repository))
                continue
            except RepositoryError:
                pass

            raise RepositoryError(f"Unknown charm or package `{target}`")
```

This block of code is executed when a target is provided - e.g. `just repo unit slurm-ops` or `just repo unit slurmctld` - to determine if a target is a `Charm` or `Package`. Since it's possible to only provide package names as targets to _repository.py_, we need to check in commands like `just repo stage slurm-ops` if any charm targets were provided as `main_cli` will still be happy and pass down the context object if only package names are provided, but charm-specific commands will be unhappy.

repository.py
NucciTheBoss: Updated `Repository` to use `public`/`private` terminology.

repository.py
jedel1043: > Hmm... aren't these values largely static? They're only computed at the beginning of the script, and they don't get modified at any point after the initial Repository object is created and handed off to context.

They are static, but we could still forget to modify either of them in the initialization phase, so I would err on the safe side and just add simple `packages` and `libraries` methods that compute the list of all the packages/libraries

NucciTheBoss: Cool cool. We'll just scope the properties to concatenating the lists rather than lazily loading the packages/libraries from disk on request.

I also had to update the annotations for the attributes because `Collection` does not support the `__add__` magic method. `list` can be casted down to `Collection`.

NucciTheBoss:
@jedel1043 alright my man, I've addressed your initial round of feedback here so this PR is ready for round 2. 

I like the `public`/`private` terminology better, and I agree that we should err on the side of caution for the `packages` and `libraries` properties. I thought lazy loading might be nice since not every subcommand requires the charm object to be fully loaded, but that's an over-optimization for our current needs. Let know if there's any other changes you think my patches to _repository.py_ need!

repository.py
```diff
@@ -742,11 +832,21 @@ def unit_test_cli(
...
 ):
     """Build all the specified charms."""
+    if len(charms) == 0:
+        raise RepositoryError("Please provide charms to generate Charmhub tokens for")
+
+    if len(packages) > 0:
+        logger.warning(
+            "Packages do not need to be built. Ignoring %s",
+            ", ".join(pkg.name for pkg in packages),
+        )
```
jedel1043, line 848: Thought: packages do need to be built; we call `uv build` on them to generate wheels so that the charms can import them. Though I'm not sure how useful that is for development, so I'm not against not supporting them in the build command.

jedel1043 (approved):
Looks good! I had a small thought about the build command, but that can easily be a future enhancement.

repository.py
NucciTheBoss: Hmm... they can... but I agree with your comment that it's not immediately clear how useful this functionality is. The reason I didn't include support for them in the `build_cli` function is because they're already built as part of the charm build process, and I primarily associate `just repo build` with build charms that I can then inject into my multipass instances.

For packages I'd just prefer to call `uv` directly, or have a dedicated command in _repository.py_ that facilitates public package building and publishing with `uv build` and `uv publish`. That said, this is a potential optimization for in the coming weeks once we start actually publishing the Slurm charm interfaces to PyPI.

Issue #204, opened 2026-04-08
slurmd failing `scontrol reconfigure` due to `error: Unable to initialize cgroup plugin`


dsloanm:
### Bug Description

In a Charmed HPC deployment on LXD backing cloud, running `scontrol reconfigure` on a node currently results in `slurmd.service` failing to reload its configuration. The service logs lists cgroup errors as the cause.

This results in breakage of any operation that requires a cluster-wide configuration reload using `scontrol reconfigure`.

The expected behavior is the `scontrol reconfigure` results in a successful reload of configuration on all running `slurmd.services` across all compute nodes.

### To Reproduce

On an LXD Juju controller, deploy a slurmctld controller and slurmd compute node:

```
juju add-model charmed-hpc
juju deploy slurmctld --base "ubuntu@24.04" --channel "latest/edge"
juju deploy slurmd --base "ubuntu@24.04" --channel "latest/edge"
juju integrate slurmd slurmctld
```

Wait for deployment to complete:

```
$ juju status
Model        Controller      Cloud/Region         Version  SLA          Timestamp
charmed-hpc  lxd-controller  localhost/localhost  3.6.13   unsupported  19:00:29+01:00

App        Version        Status  Scale  Charm      Channel      Rev  Exposed  Message
slurmctld  25.11.4-1ppa1  active      1  slurmctld  latest/edge  142  no       primary - UP
slurmd     25.11.4-1ppa1  active      1  slurmd     latest/edge  163  no       

Unit          Workload  Agent  Machine  Public address  Ports          Message
slurmctld/0*  active    idle   0        10.200.245.167  6817,9092/tcp  primary - UP
slurmd/0*     active    idle   1        10.200.245.150  6818/tcp       

Machine  State    Address         Inst id        Base          AZ           Message
0        started  10.200.245.167  juju-71e38f-0  ubuntu@24.04  me-Adder-WS  Running
1        started  10.200.245.150  juju-71e38f-1  ubuntu@24.04  me-Adder-WS  Running
```

Run `scontrol reconfigure` on the `slurmctld` unit:

```
$ juju exec --unit slurmctld/0 -- sudo scontrol reconfigure
```

View the `slurmd.service` logs on the `slurmd` unit and confirm reconfiguration failure:

```
$ juju exec --unit slurmd/0 -- sudo journalctl -xu slurmd
[...]
[...]

Issue #203, opened 2026-04-07
slurmrestd empty auth key


dsloanm:
### Bug Description

Deploying and integrating the `slurmrestd` results in an empty, 0-byte file at `/etc/slurm/slurm.key` on the `slurmrestd` unit.

Examining the relation data with `jhack` shows `auth_key_id` is erroneously the empty string rather than the secret ID. The `slurmctld` logs show the `SlurmrestdConnectedEvent` is run, and the contents of `slurmconfig` are present, but the ID does not appear to be written to the databag successfully.

### To Reproduce

Deploy and integrate a controller, accounting database and slurmrestd itself:

```
juju add-model charmed-hpc

juju deploy slurmctld --base "ubuntu@24.04" --channel "latest/edge"
juju deploy slurmdbd --base "ubuntu@24.04" --channel "latest/edge"
juju deploy slurmrestd --base "ubuntu@24.04" --channel "latest/edge"
juju deploy mysql --channel "8.0/stable"

juju integrate slurmctld slurmdbd
juju integrate slurmctld slurmrestd
juju integrate slurmdbd mysql:database
```

Wait for the deployment to complete:

```
$ juju status
Model        Controller      Cloud/Region         Version  SLA          Timestamp
charmed-hpc  lxd-controller  localhost/localhost  3.6.13   unsupported  16:44:21+01:00

App         Version          Status  Scale  Charm       Channel      Rev  Exposed  Message
mysql       8.0.44-0ubun...  active      1  mysql       8.0/stable   444  no       
slurmctld   25.11.4-1ppa1    active      1  slurmctld   latest/edge  142  no       primary - UP
slurmdbd    25.11.4-1ppa1    active      1  slurmdbd    latest/edge  135  no       
slurmrestd  25.11.4-1ppa1    active      1  slurmrestd  latest/edge  137  no       

Unit           Workload  Agent  Machine  Public address  Ports           Message
mysql/0*       active    idle   3        10.200.245.233  3306,33060/tcp  Primary
slurmctld/0*   active    idle   0        10.200.245.144  6817,9092/tcp   primary - UP
slurmdbd/0*    active    idle   1        10.200.245.243  6819/tcp        
slurmrestd/0*  active    idle   2        10.200.245.4    6820/tcp        

Machine  State    Address         Inst id        Base          AZ           Message
0        started  10.200.245.144  juju-bcb3e5-0  ubuntu@24.04  me-Adder-WS  Running
[...]

Issue #147, opened 2025-09-12
Keys are erroneously regenerated in non-HA setup when `slurmctld` unit is removed then re-added


dsloanm:
### Bug Description

In a Charmed HPC cluster with a non-HA `slurmctld` setup, removing the `slurmctld` unit, then adding a new one, causes the Slurm auth key and JWT key to be regenerated on the new `slurmctld` unit.

This new key is not propagated to any integrated application, resulting in loss of communication between other nodes (login, compute, etc.) and the controller due to authentication failures.

A potential fix is to extend the check at https://github.com/charmed-hpc/slurm-charms/blob/edc4736/charms/slurmctld/src/charm.py#L181 to look for the auth and JWT key secrets in the model and, if found, reuse them rather than regenerating. This may require supporting changes to `slurm_ops` and `hpc-libs`.

This issue does not occur in an HA setup as the key files are held on shared storage rather than the `slurmctld` unit.

### To Reproduce

To demonstrate this with the Slurm auth key:

1. `juju deploy slurmctld --base "ubuntu@24.04" --channel "edge"`
2. `juju deploy slurmd --base "ubuntu@24.04" --channel "edge"`
3. `juju integrate slurmctld slurmd`

Wait for deployment to complete then confirm keys on the `slurmctld` unit and `slurmd` unit are the same:

5. `juju exec --unit slurmctld/0 -- sha1sum /etc/slurm/slurm.key`
```
7d705efaac47f075362ceb4cf16f195371d6b993  /etc/slurm/slurm.key
```
6. `juju exec --unit slurmd/0 -- sha1sum /etc/slurm/slurm.key`
```
7d705efaac47f075362ceb4cf16f195371d6b993  /etc/slurm/slurm.key
```

Remove `slurmctld` unit, add a new one, and confirm keys on the `slurmctld` unit and `slurmd` unit are different:

7.  `juju remove-unit slurmctld/0`
8. `juju add-unit slurmctld`
9. `juju exec --unit slurmctld/1 -- sha1sum /etc/slurm/slurm.key`
```
d180d93dd7bdf60b05679a7abc649bbdde77bd34  /etc/slurm/slurm.key
```
10.  `juju exec --unit slurmctld/0 -- sha1sum /etc/slurm/slurm.key`
```
7d705efaac47f075362ceb4cf16f195371d6b993  /etc/slurm/slurm.key
```

### Environment

`slurmctld` revision 120
`slurmd` revision 141

### Relevant log output

[...]

PR #207, opened 2026-04-09
feat: implement auth key rotation

dsloanm:
> ~Draft commit as relies on a personal fork of `hpc-libs`.~
> ~Updates to the secrets handling must be merged into an `hpc-libs` release to unblock this PR.~
> Out of draft status. Necessary secret handing changes have been included in this PR

# Pre-submission checklist

 * [X] I read and followed the CONTRIBUTING guidelines.
 * [X] I have ensured that lint, typecheck, and unit tests complete successfully.

## Summary of changes

The new `rotate-auth-key` action is implemented on `slurmctld` to perform a cluster-wide rotation of the Slurm auth key. To facilitate this:

* Slurm key manager in `slurm-ops` has been rewritten to support Slurm's multiple key setup using `slurm.jwks` in place of `slurm.key`: https://slurm.schedmd.com/authentication.html#multiple_key_setup
* SSDLC-compliant logging of key changes to `slurm.jwks` is added

Rotation is performed by:

* Admin runs `juju run slurmctld/leader rotate-auth-key`
* Controller adds a new key to the `slurm.jwks` file alongside the old key
* The auth key Juju secret is updated to a new revision with the new key
* All secret observers/other Slurm units receive a SecretChangedEvent and update local key file with the new key
* Controller receives a SecretRemoveEvent after all Slurm units have updated to the new key and removes the old key from the `slurm.jwks` file

Slurm interfaces have been updated to use a model-level auth key secret:

* A constant `AUTH_KEY_LABEL` is now used to refer to the secret
* `auth_key_id` is renamed to `auth_secret_id` in `ControllerData`
* `auth_key_id` is reused to refer to ID for the key revision rather than secret itself
* Auth key secret is no longer created in the interface. This is now done in charm code
* Auth key revisions are no longer removed when a relation is broken.
  * Secret access does not require manual revocation in charm code: ["If the relation is removed, the secret access will be revoked."](https://documentation.ubuntu.com/juju/3.6/reference/secret/#secret-lifecycle) 
  * Secret revisions do not require manual cleanup in charm code. They are cleaned up when `slurmctld` is removed:  ["juju has code which removes any charm owned secrets if the application is removed"](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$10ns2_hnyZqrj8baKHDxUkn9yGmdh9jqWqDfeZVy4gk?via=ubuntu.com&via=matrix.org&via=tchncs.de)

The rotation is currently not a zero-downtime process due to:

* https://github.com/charmed-hpc/slurm-charms/issues/204 preventing reload of slurmd.service, necessitating a restart to load the new key
* slurmdbd.service not reloading its auth key on a `systemctl reload slurmdbd.service`, necessitating a restart.
  * This may be a Slurm bug. It is inconsistent with the behavior of other Slurm services (`systemctl reload sackd.service` reloads the key as expected) and documentation states the `SIGHUP` signal ["Reloads the slurm configuration files"](https://slurm.schedmd.com/slurmdbd.html#OPT_SIGHUP)
* slurmrestd.service not supporting a reload signal, necessitating a restart. The process exits on a `systemctl reload slurmrestd.service`. No relevant signals are listed in https://slurm.schedmd.com/slurmrestd.html#SECTION_SIGNALS

All workarounds in the code are marked with `#TODO` or `#FIXME` comments.

This PR fixes:

* https://github.com/charmed-hpc/slurm-charms/issues/147
* https://github.com/charmed-hpc/slurm-charms/issues/203

## Docs

* [X] I have created a pull request to add or update relevant documentation in [charmed-hpc/docs](https://github.com/charmed-hpc/docs) or another documentation location.

Or:

* [ ] I confirm that this pull request requires no changes or additions to documentation.

https://github.com/charmed-hpc/docs/pull/130

### DIFF
--- a/charms/sackd/src/charm.py
+++ b/charms/sackd/src/charm.py
@@ -16,11 +16,10 @@
 """Charmed operator for `sackd`, Slurm's authentication kiosk service."""
 
 import logging
-from subprocess import CalledProcessError
 
 import ops
-from charmed_hpc_libs.ops import call
-from charmed_hpc_libs.ops.conditions import StopCharm, block_unless, refresh, wait_unless
+from charmed_hpc_libs.errors import SystemdError
+from charmed_hpc_libs.ops import StopCharm, block_unless, refresh, systemctl, wait_unless
 from charmed_slurm_sackd_interface import (
     AUTH_KEY_LABEL,
     SackdProvider,
@@ -147,8 +146,8 @@ class SackdCharm(ops.CharmBase):
         # Necessary to load new key from file into the service
         # TODO: replace with self.service.reload()
         try:
-            call("/usr/bin/systemctl", "reload", "sackd.service")
-        except CalledProcessError as e:
+            systemctl("reload", "sackd.service")
+        except SystemdError as e:
             logger.error("failed to reload sackd.service. reason:\n%s", e)
             event.defer()
             raise StopCharm(
--- a/charms/sackd/src/charm.py
+++ b/charms/sackd/src/charm.py
@@ -16,6 +16,7 @@
 """Charmed operator for `sackd`, Slurm's authentication kiosk service."""
 
 import logging
+from subprocess import CalledProcessError
 
 import ops
 from charmed_hpc_libs.ops.conditions import StopCharm, block_unless, refresh, wait_unless
@@ -43,6 +44,7 @@ class SackdCharm(ops.CharmBase):
         self.sackd = SackdManager(snap=False)
         framework.observe(self.on.install, self._on_install)
         framework.observe(self.on.update_status, self._on_update_status)
+        framework.observe(self.on.secret_changed, self._on_secret_changed)
 
         self.slurmctld = SackdProvider(self, SACKD_INTEGRATION_NAME)
         framework.observe(
@@ -90,7 +92,7 @@ class SackdCharm(ops.CharmBase):
         data = self.slurmctld.get_controller_data(event.relation.id)
 
         try:
-            self.sackd.key.set(data.auth_key)
+            self.sackd.key.set(data.auth_key, data.auth_key_content_id)
             self.sackd.conf_server = data.controllers
             self.sackd.service.enable()
             self.sackd.service.restart()
@@ -114,6 +116,44 @@ class SackdCharm(ops.CharmBase):
             event.defer()
             raise StopCharm(
                 ops.BlockedStatus("Failed to stop `sackd`. See `juju debug-log` for details")
+            )
+
+    @refresh
+    @block_unless(sackd_installed)
+    def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
+        """Handle when a secret is changed."""
+        if event.secret.label != AUTH_KEY_LABEL:
+            logger.warning("secret with label '%s' changed. ignoring", event.secret.label)
+            return
+
+        content = event.secret.get_content(refresh=True)
+        auth_key = content.get("key")
+        auth_key_id = content.get("keyid")
+        if not auth_key or not auth_key_id:
+            logger.error(
+                "auth key or key ID is empty in secret with label '%s'", event.secret.label
+            )
+            event.defer()
+            raise StopCharm(
+                ops.BlockedStatus(
+                    "Failed to retrieve Slurm authentication key. See `juju debug-log` for details"
+                )
+            )
+
+        self.sackd.key.set(auth_key, auth_key_id)
+
+        # Necessary to load new key from file into the service
+        # TODO: replace with self.service.reload()
+        slurm_service = "sackd.service"
+        try:
+            call("/usr/bin/systemctl", "reload", slurm_service)
+        except CalledProcessError as e:
+            logger.exception("failed to reload %s. reason:\n%s", slurm_service, e)
+            event.defer()
+            raise StopCharm(
+                ops.BlockedStatus(
+                    "Failed to reload %s. See `juju debug-log` for details" % slurm_service
+                )
             )
 
 
--- a/charms/sackd/src/charm.py
+++ b/charms/sackd/src/charm.py
@@ -146,15 +146,14 @@ class SackdCharm(ops.CharmBase):
 
         # Necessary to load new key from file into the service
         # TODO: replace with self.service.reload()
-        slurm_service = "sackd.service"
         try:
-            call("/usr/bin/systemctl", "reload", slurm_service)
+            call("/usr/bin/systemctl", "reload", "sackd.service")
         except CalledProcessError as e:
-            logger.exception("failed to reload %s. reason:\n%s", slurm_service, e)
+            logger.error("failed to reload sackd.service. reason:\n%s", e)
             event.defer()
             raise StopCharm(
                 ops.BlockedStatus(
-                    "Failed to reload %s. See `juju debug-log` for details" % slurm_service
+                    "Failed to reload sackd.service. See `juju debug-log` for details"
                 )
             )
 
--- a/charms/sackd/src/charm.py
+++ b/charms/sackd/src/charm.py
@@ -152,7 +152,7 @@ class SackdCharm(ops.CharmBase):
             event.defer()
             raise StopCharm(
                 ops.BlockedStatus(
-                    "Failed to reload sackd.service. See `juju debug-log` for details"
+                    "Failed to reload `sackd` configuration. See `juju debug-log` for details"
                 )
             )
 
--- a/charms/sackd/src/charm.py
+++ b/charms/sackd/src/charm.py
@@ -19,8 +19,10 @@ import logging
 from subprocess import CalledProcessError
 
 import ops
+from charmed_hpc_libs.ops import call
 from charmed_hpc_libs.ops.conditions import StopCharm, block_unless, refresh, wait_unless
 from charmed_slurm_sackd_interface import (
+    AUTH_KEY_LABEL,
     SackdProvider,
     SlurmctldDisconnectedEvent,
     SlurmctldReadyEvent,
@@ -92,7 +94,7 @@ class SackdCharm(ops.CharmBase):
         data = self.slurmctld.get_controller_data(event.relation.id)
 
         try:
-            self.sackd.key.set(data.auth_key, data.auth_key_content_id)
+            self.sackd.key.set(data.auth_key, data.auth_key_id)
             self.sackd.conf_server = data.controllers
             self.sackd.service.enable()
             self.sackd.service.restart()
--- a/charms/slurmctld/src/charm.py
+++ b/charms/slurmctld/src/charm.py
@@ -18,6 +18,7 @@
 
 import logging
 import secrets
+from uuid import uuid4
 
 import mail
 import ops
@@ -138,6 +139,9 @@ class SlurmctldCharm(ops.CharmBase):
         framework.observe(self.on.start, self._on_start)
         framework.observe(self.on.config_changed, self._on_config_changed)
         framework.observe(self.on.update_status, self._on_update_status)
+        framework.observe(self.on.secret_changed, self._on_secret_changed)
+        framework.observe(self.on.secret_remove, self._on_secret_remove)
+        framework.observe(self.on.rotate_auth_key_action, self._on_rotate_auth_key_action)
         framework.observe(self.on.show_current_config_action, self._on_show_current_config_action)
         framework.observe(self.on.set_node_state_action, self._on_set_node_state_action)
 
@@ -217,8 +221,24 @@ class SlurmctldCharm(ops.CharmBase):
                 # new unit is elected leader as it is being deployed
                 if not self.slurmctld.jwt.path.exists():
                     self.slurmctld.jwt.generate()
-                if not self.slurmctld.key.path.exists():
-                    self.slurmctld.key.generate()
+
+                try:
+                    secret = self.model.get_secret(label=AUTH_KEY_LABEL)
+                    logger.debug("auth key secret found. skipping generation")
+
+                    if not self.slurmctld.key.path.exists():
+                        logger.warning("auth key file not found. restoring from secret")
+                        content = secret.get_content()
+                        self.slurmctld.key.set(content["key"], content["keyid"])
+                except ops.SecretNotFoundError:
+                    key = self.slurmctld.key.generate()
+                    # Auth key entries must each have a unique ID consistent across all units.
+                    # Secret revision number cannot be used as it is not known to secret observers.
+                    # Secret `unique_identifier` property is not unique per revision.
+                    # Therefore, a newly generated UUID is included with the secret contents.
+                    key_id = str(uuid4())
+                    self.app.add_secret({"key": key, "keyid": key_id}, label=AUTH_KEY_LABEL)
+                    self.slurmctld.key.set(key, key_id)
 
             self.unit.set_workload_version(self.slurmctld.version())
         except SlurmOpsError as e:
@@ -401,10 +421,11 @@ class SlurmctldCharm(ops.CharmBase):
     @block_unless(slurmctld_installed)
     def _on_sackd_connected(self, event: SackdConnectedEvent) -> None:
         """Handle when a new `sackd` application is connected."""
+        auth_key_id = self.model.get_secret(label=AUTH_KEY_LABEL).get_info().id
         new_endpoints = [f"{c}:{SLURMCTLD_PORT}" for c in self._get_controllers()]
         self.sackd.set_controller_data(
             ControllerData(
-                auth_key=self.slurmctld.key.get(),
+                auth_key_id=auth_key_id,
                 controllers=new_endpoints,
             ),
             integration_id=event.relation.id,
@@ -415,6 +436,7 @@ class SlurmctldCharm(ops.CharmBase):
     @block_unless(slurmctld_installed)
     def _on_slurmd_ready(self, event: SlurmdReadyEvent) -> None:
         """Handle when partition data is ready from a `slurmd` application."""
+        auth_key_id = self.model.get_secret(label=AUTH_KEY_LABEL).get_info().id
         data = self.slurmd.get_compute_data(event.relation.id)
         name = data.partition.partition_name
         include = f"slurm.conf.{name}"
@@ -435,7 +457,7 @@ class SlurmctldCharm(ops.CharmBase):
         new_endpoints = [f"{c}:{SLURMCTLD_PORT}" for c in self._get_controllers()]
         self.slurmd.set_controller_data(
             ControllerData(
-                auth_key=self.slurmctld.key.get(),
+                auth_key_id=auth_key_id,
                 controllers=new_endpoints,
             ),
             integration_id=event.relation.id,
@@ -463,9 +485,10 @@ class SlurmctldCharm(ops.CharmBase):
     @block_unless(slurmctld_installed)
     def _on_slurmdbd_connected(self, event: SlurmdbdConnectedEvent) -> None:
         """Handle when a new `slurmdbd` application is connected."""
+        auth_key_id = self.model.get_secret(label=AUTH_KEY_LABEL).get_info().id
         self.slurmdbd.set_controller_data(
             ControllerData(
-                auth_key=self.slurmctld.key.get(),
+                auth_key_id=auth_key_id,
                 jwt_key=self.slurmctld.jwt.get(),
             ),
             integration_id=event.relation.id,
@@ -520,9 +543,10 @@ class SlurmctldCharm(ops.CharmBase):
     @block_unless(slurmctld_installed)
     def _on_slurmrestd_connected(self, event: SlurmrestdConnectedEvent) -> None:
         """Handle when a new `slurmrestd` application is connected."""
+        auth_key_id = self.model.get_secret(label=AUTH_KEY_LABEL).get_info().id
         self.slurmrestd.set_controller_data(
             ControllerData(
-                auth_key=self.slurmctld.key.get(),
+                auth_key_id=auth_key_id,
                 slurmconfig={
                     "slurm.conf": self.slurmctld.config.load(),
                     **{k: v.load() for k, v in self.slurmctld.config.includes.items()},
@@ -625,6 +649,65 @@ class SlurmctldCharm(ops.CharmBase):
         logger.info("`%s` configuration deleted successfully", self.slurmctld.oci.name)
 
         self._reconfigure()
+
+    @block_unless(slurmctld_installed)
+    def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
+        """Handle when a secret is changed."""
+        if event.secret.label != AUTH_KEY_LABEL:
+            logger.warning("secret with label '%s' changed. ignoring", event.secret.label)
+            return
+
+        # Force tracking of the new revision. Needed as this application is both the owner and
+        # an observer. Without this get_content call, the secret-remove event is not emitted
+        # after all other observers complete their key rotation, as this unit, and backup units in
+        # an HA configuration, still observe the old revision
+        # TODO: Confirm if this behavior has changed in Juju 4
+        event.secret.get_content(refresh=True)
+
+    @leader
+    @refresh
+    @block_unless(slurmctld_installed)
+    def _on_secret_remove(self, event: ops.SecretRemoveEvent) -> None:
+        """Handle when a secret is removed."""
+        if event.secret.label != AUTH_KEY_LABEL:
+            logger.warning("secret with label '%s' removed. ignoring", event.secret.label)
+            return
+
+        self.slurmctld.key.keep_latest_key()
+        # Reconfigure must come before revision is removed to ensure secret ID can be retrieved for
+        # slurmrestd databag. This prevents a SecretNotFoundError.
+        # TODO: This will not be necessary once merging of values into the databag is implemented
+        # and the secret ID no longer needs set in _reconfigure.
+        self._reconfigure()
+
+        event.remove_revision()
+
+    @refresh
+    def _on_rotate_auth_key_action(self, event: ops.ActionEvent) -> None:
+        """Rotate the Slurm authentication key across the cluster."""
+        if not self.unit.is_leader():
+            event.fail("Only the leader unit can rotate the authentication key.")
+            return
+
+        # Update secrets backend with new key revision
+        new_key = self.slurmctld.key.generate()
+        new_key_id = str(uuid4())
+        try:
+            # Update key file first to ensure key is available before observers apply new revision
+            self.slurmctld.key.add(new_key, new_key_id)
+            self._reconfigure()
+        except Exception as e:
+            logger.error("failed to update auth key. reason:\n%s", e)
+            event.fail("Failed to update auth key. See `juju debug-log` for details.")
+            return
+
+        try:
+            secret = self.model.get_secret(label=AUTH_KEY_LABEL)
+            secret.set_content({"key": new_key, "keyid": new_key_id})
+        except (ops.SecretNotFoundError, ModelError) as e:
+            logger.error("failed to publish new auth key secret. reason:\n%s", e)
+            event.fail("Failed to publish new auth key secret. See `juju debug-log` for details.")
+            return
 
     def _on_show_current_config_action(self, event: ops.ActionEvent) -> None:
         """Show current slurm.conf."""
@@ -822,14 +905,21 @@ class SlurmctldCharm(ops.CharmBase):
             )
 
         if self.slurmrestd.is_joined():
-            self.slurmrestd.set_controller_data(
-                ControllerData(
-                    slurmconfig={
-                        "slurm.conf": self.slurmctld.config.load(),
-                        **{k: v.load() for k, v in self.slurmctld.config.includes.items()},
-                    }
+            # Workaround for: https://github.com/charmed-hpc/slurm-charms/issues/203
+            # TODO: Remove setting of key ID once merging of databag info is implemented. Only the
+            # slurmconfig needs updated. The auth_key_id should not be overwritten
+            auth_key_id = self.model.get_secret(label=AUTH_KEY_LABEL).get_info().id
+            for integration in self.model.relations.get(SLURMRESTD_INTEGRATION_NAME, []):
+                self.slurmrestd.set_controller_data(
+                    ControllerData(
+                        auth_key_id=auth_key_id,
+                        slurmconfig={
+                            "slurm.conf": self.slurmctld.config.load(),
+                            **{k: v.load() for k, v in self.slurmctld.config.includes.items()},
+                        },
+                    ),
+                    integration_id=integration.id,
                 )
-            )
 
     def _merge_controller_data(self, app: SackdRequirer | SlurmdRequirer, new_endpoints) -> None:
         """Merge new controller endpoints with existing controller data."""
--- a/charms/slurmctld/src/charm.py
+++ b/charms/slurmctld/src/charm.py
@@ -690,7 +690,7 @@ class SlurmctldCharm(ops.CharmBase):
             # Update key file first to ensure key is available before observers apply new revision
             self.slurmctld.key.add(new_key, new_key_id)
             self._reconfigure()
-        except Exception as e:
+        except (SlurmOpsError, ValueError) as e:
             logger.error("failed to update auth key. reason:\n%s", e)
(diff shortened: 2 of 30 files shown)


dsloanm:
Test failure due to:

```
Run sudo snap install just --classic
error: cannot install "just": cannot query the store for updates:
       got unexpected HTTP status code 503 via POST to
       "https://api.snapcraft.io/v2/snaps/refresh"
Error: Process completed with exit code 1.
```

Will rerun later.

charms/slurmctld/src/charm.py
```diff
@@ -614,6 +634,59 @@ def _on_oci_runtime_disconnected(self, _: OCIRuntimeDisconnectedEvent) -> None:
...
+            logger.warning("secret with label '%s' changed. ignoring", event.secret.label)
+            return
+
+        # Force tracking of the new revision. Needed as this application is both the owner and
+        # an observer. Without this get_content call, the secret-remove event is not emitted
+        # after all other observers complete their key rotation, as this unit, and backup units in
+        # an HA configuration, still observe the old revision
+        # TODO: Confirm if this behavior has changed in Juju 4
+        event.secret.get_content(refresh=True)
+
```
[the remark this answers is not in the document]
dsloanm: This is intentional. Backup `slurmctld` units share `slurm.jwks` with the primary via a shared file system.

pyproject.toml
```diff
@@ -73,7 +73,7 @@ members = [
 [tool.uv.sources]
 slurm-ops = { workspace = true }
 ubuntu-drivers-common = { git = "https://github.com/canonical/ubuntu-drivers-common", rev = "554b91edfd3699625dbed90f679abb31a897b76e" }
-hpc-libs = { git = "https://github.com/charmed-hpc/hpc-libs", rev = "9fba44f1729eba44be5577f679e5c03f588bcfff" }
+hpc-libs = { git = "https://github.com/dsloanm/hpc-libs", branch="auth-key" }
```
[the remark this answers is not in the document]
dsloanm: This is intentional and the reason for the PR being in draft status.

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -588,3 +696,32 @@ def _save_options(self, options: Mapping[str, Any]) -> None:
...
+    # https://github.com/canonical/operator/blob/9affc1e/ops/log.py#L154
+    log_message = {
+        "datetime": datetime.datetime.now().isoformat(),
+        "level": level,
+        "type": "security",
+        "appid": "slurm-charms",
+        "event": f"{event_type}:{event_data}",
+        "description": description,
+    }
+    _logger.debug(json.dumps(log_message))
```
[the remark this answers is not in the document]
dsloanm: Valid - will fix.

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,140 @@ def path(self) -> Path:
...
+            "kty": "oct",
+            "kid": key_id,
+            "k": key,
+        }
+
+    def _read_jwks(self) -> dict[str, list[dict[str, str]]]:
+        try:
+            data = json.loads(self._file.read_text())
+        except FileNotFoundError:
+            data = {"keys": []}
```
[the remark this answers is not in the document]
dsloanm: Valid - will include additional validation and update the docstring.

charms/slurmctld/src/charm.py
```diff
@@ -203,8 +208,19 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+                    logger.debug("auth key secret found. skipping generation")
+                except ops.SecretNotFoundError:
+                    key = self.slurmctld.key.generate()
+                    # Auth key entries must each have a unique ID consistent across all units.
+                    # Secret revision number cannot be used as it is not known to secret observers.
+                    # Secret `unique_identifier` property is not unique per revision.
+                    # Therefore, a newly generated UUID is included with the secret contents.
+                    key_id = str(uuid4())
+                    self.app.add_secret({"key": key, "keyid": key_id}, label=AUTH_KEY_LABEL)
+                    self.slurmctld.key.add(key, key_id)
```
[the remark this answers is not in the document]
dsloanm: Upgrade scenario from slurm.key does not need to be considered.

Accounting for when the secret exists but the file does not is valid. Will fix.

charms/slurmctld/src/charm.py
```diff
@@ -614,6 +634,59 @@ def _on_oci_runtime_disconnected(self, _: OCIRuntimeDisconnectedEvent) -> None:
...
+            secret = self.model.get_secret(label=AUTH_KEY_LABEL)
+            secret.set_content({"key": new_key, "keyid": new_key_id})
+        except (ops.SecretNotFoundError, ModelError) as e:
+            logger.error("failed to rotate auth key. reason:\n%s", e)
+            event.fail("Failed to rotate auth key. See `juju debug-log` for details.")
+            return
+
+        self.slurmctld.key.add(new_key, new_key_id)
+        self._reconfigure()
+
```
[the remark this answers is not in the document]
dsloanm: Valid - will reorder

charms/slurmctld/src/state.py
```diff
@@ -151,6 +152,26 @@ def slurmctld_ready(charm: "SlurmctldCharm") -> bool:
...
+    except OSError:
+        return ConditionEvaluation(False, "Authentication key file inaccessible")
+
+    if len(keys.get("keys", [])) > 1:
+        return ConditionEvaluation(
+            False,
+            "Authentication key rotation in progress. Waiting for rotation to complete.",
+        )
+
+    return ConditionEvaluation(True, "")
```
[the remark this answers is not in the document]
dsloanm: Valid - will also use the new validation in the key reading function to replace these checks.

dsloanm:
Co-pilot comments addressed

dsloanm:
Tests failed due to:
```
 Installing woke ... https://github.com/get-woke/woke
  get-woke/woke info checking GitHub for latest tag
  get-woke/woke info found version: 0.19.0 for v0.19.0/linux/amd64
  Error: Process completed with exit code 1.
```
Will rerun

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
+        )
 
     @property
     def path(self) -> Path:
-        """Get the path to the `slurm.key` secret file."""
+        """Get the path to the `slurm.jwks` secret file.
+
+        Returns:
+            Path to the `slurm.jwks` file.
+        """
```
NucciTheBoss: [nit]: This `Returns` section is redundant here. The description already captures that the path is returned by this property.

```suggestion
        """Get the path to the `slurm.jwks` secret file."""
```

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
-    def set(self, secret: str) -> None:
-        """Set the contents of the `slurm.key` secret file."""
-        self._file.write_bytes(base64.b64decode(secret.encode()))
-        self._file.chmod(0o600)
-        shutil.chown(self._file, self._user, self._group)
+        Returns:
+            A base64-encoded random byte string suitable for use as a Slurm authentication key.
+        """
+        key = base64.b64encode(secrets.token_bytes(2048)).decode()
+        return key
```
NucciTheBoss: We don't need to instantiate the variable `key` here if we're then immediately returning its value to the outer scope:

```suggestion
        return base64.b64encode(secrets.token_bytes(2048)).decode()
```

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
-        """Get the path to the `slurm.key` secret file."""
+        """Get the path to the `slurm.jwks` secret file.
+
+        Returns:
+            Path to the `slurm.jwks` file.
+        """
         return self._file
 
+    @staticmethod
+    def _get_new_entry(key: str, key_id: str) -> dict[str, str]:
```
NucciTheBoss: How about `_build_new_entry` instead of `_get_new_entry`? `get` implies to me that we're pulling the value from somewhere else when really we're building a new key entry for the _slurm.jwks_ file. I found `_get_new_entry` confusing initially because I thought we were getting the key from somewhere when actually build a new key:


```suggestion
    def _build_new_entry(key: str, key_id: str) -> dict[str, str]:
```

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
+
+        Args:
+            key: The base64-encoded key material to add.
+            key_id: A unique identifier for the key.
+
+        Raises:
+            ValueError: If `key` or `key_id` is empty.
+        """
+        data = self.get()
+        new_entry = self._get_new_entry(key, key_id)
```
NucciTheBoss: Should the instantiation of `new_entry` come before the instantiation of `data`? Why go through all the trouble of loading the _slurm.jwks_ and validating if it's correct when we already know this method will fail because the caller passed bad inputs from the outer scope (empty strings):

```suggestion
        new_entry = self._get_new_entry(key, key_id)
        data = self.get()
```

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
+        except OSError as e:
+            raise SlurmOpsError(f"Failed to read {self._file}") from e
+        except json.JSONDecodeError as e:
+            raise SlurmOpsError(f"Failed to parse {self._file}") from e
+
+        if not isinstance(data, dict):
+            raise SlurmOpsError(f"Invalid key data in {self._file}: expected a JSON object")
+
+        if not isinstance(data.get("keys"), list):
+            raise SlurmOpsError(f"Invalid key data in {self._file}: 'keys' must be a list")
```
NucciTheBoss, line 597: [foresight]: this feels like it could be wrapped nicely under a `JWKSFile` pydantic model or something similar. This way we don't have to call `isinstance` directly, we can fully validate the shape of the parsed _slurm.jwks_ file - e.g. no unexpected values for the `alg` or `kty` fields, and we can simplify some of our annotations.

That said, this is an optimization that can be saved as we prepare for our stable release later this year as things may change or we'll have new insights after adding the JWT rotation code.

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
     def path(self) -> Path:
-        """Get the path to the `slurm.key` secret file."""
+        """Get the path to the `slurm.jwks` secret file.
+
+        Returns:
+            Path to the `slurm.jwks` file.
+        """
         return self._file
 
+    @staticmethod
```
NucciTheBoss, line 636: [nit]: private `staticmethod` definitions should come after private method definition. E.g. change the class definition order so that `_write` comes before this `staticmethod`.

tests/integration/test_charm.py
```diff
@@ -228,6 +228,67 @@ def test_set_node_state(juju: jubilant.Juju) -> None:
...
+    juju.wait(
+        lambda status: jubilant.all_active(status, *SLURM_APPS),
+        error=lambda status: jubilant.any_error(status, *SLURM_APPS),
+    )
+
+    # Check authentication key has been updated on all units
+    # Key rotation does not complete until the secret-remove event has completed on slurmctld. This
+    # event is triggered after all observers have updated to the new secret revision. There may be
+    # a gap where all Slurm applications are in ActiveStatus but secret-remove has not yet run so
+    # the old key is still present on slurmctld. Account for this with tenacity
```
NucciTheBoss: [nit]: missing period.

```suggestion
    # the old key is still present on slurmctld. Account for this with tenacity.
```

tests/integration/test_charm.py
```diff
@@ -228,6 +228,67 @@ def test_set_node_state(juju: jubilant.Juju) -> None:
...
+                key_entry = json.loads(result.stdout)
+                assert len(key_entry["keys"]) == 1
+                assert key_entry == new_key_entry, f"auth key rotation failed on: {unit}"
+
+            # Check units can communicate with controller
+            assert juju.exec("sinfo", unit=sackd_unit).success
+            assert juju.exec("sinfo", unit=slurmd_unit).success
+            # Database does not have client tools installed. Query from the controller
+            assert juju.exec("sacct", unit=slurmctld_unit).success
+            # TODO: confirm slurmrestd connectivity
```
NucciTheBoss, line 288: Why didn't you include tests for confirming slurmrestd connectivity here?

tests/integration/test_charm.py
```diff
@@ -228,6 +228,67 @@ def test_set_node_state(juju: jubilant.Juju) -> None:
...
+
+    logger.info("testing that the `rotate-auth-key` action updates the Slurm authentication key")
+
+    # Gather existing authentication key from `slurmctld/0`.
+    result = juju.exec("sudo cat /etc/slurm/slurm.jwks", unit=slurmctld_unit)
+    initial_key_entry = json.loads(result.stdout)
+
+    juju.run(slurmctld_unit, "rotate-auth-key")
+
+    # Wait for action to complete and for all Slurm applications to return to ActiveStatus
```
NucciTheBoss: [nit]: missing period.

```suggestion
    # Wait for action to complete and for all Slurm applications to return to ActiveStatus.
```

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -588,3 +716,32 @@ def _save_options(self, options: Mapping[str, Any]) -> None:
         self._env_manager.set(
             {f"{self._service.upper()}_OPTIONS": marshal_options(options)}, quote=False
         )
+
+
+def _log_security_event(level: str, event_type: str, event_data: str, description: str) -> None:
```
NucciTheBoss, line 727: [foresight]: we might want to eventually add this function to `charmed-hpc-libs`. IIRC currently the only charms we have that require this method right now are Slurm, but I could see us adding this function to the SSSD charm in the future since it's configuration file also contains sensitive secrets like LDAP server bind passwords.

Needs more thoughts and probably a specification however.

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
+            key_id: A unique identifier for the key.
+
+        Returns:
+            A dictionary representing a JWK entry with `alg`, `kty`, `kid`, and `k` fields,
+            as described in the Slurm authentication documentation.
+
+        Raises:
+            ValueError: If `key` or `key_id` is empty.
+
+        Note:
```
NucciTheBoss: [nit]: Should we `See Also` instead? Also, _I think_ the Google Docstring format uses "Notes" instead of "Note", but I could be mistaken.

```suggestion
        See Also:
```

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
+        if not key:
+            raise ValueError("Empty key provided")
+        if not key_id:
+            raise ValueError("Empty key ID provided")
+        return {
+            "alg": "HS256",
+            "kty": "oct",
+            "kid": key_id,
+            "k": key,
+        }
```
NucciTheBoss, line 664: [foresight]: This is also a good candidate for a pydantic model since we have specific fields that we know need to be here.

charms/slurmctld/src/charm.py
```diff
@@ -217,8 +222,24 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+                        content = secret.get_content()
+                        self.slurmctld.key.set(content["key"], content["keyid"])
+                except ops.SecretNotFoundError:
+                    key = self.slurmctld.key.generate()
+                    # Auth key entries must each have a unique ID consistent across all units.
+                    # Secret revision number cannot be used as it is not known to secret observers.
+                    # Secret `unique_identifier` property is not unique per revision.
+                    # Therefore, a newly generated UUID is included with the secret contents.
+                    key_id = str(uuid4())
+                    self.app.add_secret({"key": key, "keyid": key_id}, label=AUTH_KEY_LABEL)
```
NucciTheBoss, line 235: Is there a particular reason you opted for using `keyid` instead of `key_id` in the secret content? The reasoning for my gripe is that everywhere in the charm code, we refer to the value as `*_key_id` and not `*_keyid`. I could see myself easily stumbling over this inconsistency when working with the secrets' content as I'm used to working with the `*_key_id` naming scheme.

If this is because of a limitation with Juju Secrets, I'll learn to live with the pain, but we should try to be consistent with how we refer to the key ID value.

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +503,160 @@ def path(self) -> Path:
...
     def __init__(self, ops_manager: OpsManager, /, user: str, group: str) -> None:
-        self._file = ops_manager.etc_path / "slurm.key"
+        self._file = ops_manager.etc_path / "slurm.jwks"
         self._user = user
         self._group = group
 
-    def get(self) -> str:
-        """Get the contents of the current `slurm.key` secret file."""
-        return base64.b64encode(self._file.read_bytes()).decode()
+    def generate(self) -> str:
```
NucciTheBoss: Since `_SlurmSecretManager` no longer inherits from `SecretManager`, what if we had `generate` return a tuple value instead of just string. This way when we call `generate`, we get the complete value we need - key content and key id - with a single call instead of first calling `generate` and the the uuid module directly:

```python
def generate(self) -> tuple[str, str]:
    return base64.b64encode(secrets.token_bytes(2048)).decode(), str(uuid4())
```

and then:

```python
key, key_id = self.slurmctld.key.generate()
```

Imo this is cleaner than calling `uuid4` directly in the event handler, and it ensures that `generate` provides us with the complete value that we need. Also, on the plus side, if we ever change how we generate the unique key ids, we don't have to hunt all across the 9 realms of Skyrim to squash all the places where we're calling `uuid4` directly.

charms/sackd/src/charm.py
```diff
@@ -116,6 +120,44 @@ def _on_slurmctld_disconnected(self, event: SlurmctldDisconnectedEvent) -> None:
...
+
+        self.sackd.key.set(auth_key, auth_key_id)
+
+        # Necessary to load new key from file into the service
+        # TODO: replace with self.service.reload()
+        slurm_service = "sackd.service"
+        try:
+            call("/usr/bin/systemctl", "reload", slurm_service)
+        except CalledProcessError as e:
+            logger.exception("failed to reload %s. reason:\n%s", slurm_service, e)
```
NucciTheBoss: Is there a particular reason you opted for `logger.exception()` over `logger.error()` here. We generally use `logger.error()` in the charms currently, but I can see merits for using `logger.exception()`. That said, I think we should aim for consistency before introducing a logging change like this.

charms/sackd/src/charm.py
```diff
@@ -116,6 +120,44 @@ def _on_slurmctld_disconnected(self, event: SlurmctldDisconnectedEvent) -> None:
...
+            event.defer()
+            raise StopCharm(
+                ops.BlockedStatus(
+                    "Failed to reload %s. See `juju debug-log` for details" % slurm_service
```
NucciTheBoss: Using the old-style percent formatting here goes against modern Python development standards. Unless we're interfacing directly with the logging module which has it's own formatting system, we should use f-strings which is the modern development standard:


```suggestion
                    "Failed to reload `sackd`. See `juju debug-log` for details"
```

charms/sackd/src/charm.py
```diff
@@ -116,6 +120,44 @@ def _on_slurmctld_disconnected(self, event: SlurmctldDisconnectedEvent) -> None:
...
+                )
+            )
+
+        self.sackd.key.set(auth_key, auth_key_id)
+
+        # Necessary to load new key from file into the service
+        # TODO: replace with self.service.reload()
+        slurm_service = "sackd.service"
+        try:
+            call("/usr/bin/systemctl", "reload", slurm_service)
```
NucciTheBoss: I would just call `sackd.service` directly since we already have a TODO item here to replace this mechanism with `self.sackd.service.reload()`:

```suggestion
        # Necessary to load new key from file into the service
        # TODO: replace with self.service.reload()
        try:
            call("/usr/bin/systemctl", "reload", "sackd.service")
```

I'd rather go with the Occam's Razor approach here since, well, we know sackd is the only Slurm service we're interacting with here 😅

charms/slurmctld/src/charm.py
```diff
@@ -626,6 +651,65 @@ def _on_oci_runtime_disconnected(self, _: OCIRuntimeDisconnectedEvent) -> None:
...
+            return
+
+        # Update secrets backend with new key revision
+        new_key = self.slurmctld.key.generate()
+        new_key_id = str(uuid4())
+        try:
+            # Update key file first to ensure key is available before observers apply new revision
+            self.slurmctld.key.add(new_key, new_key_id)
+            self._reconfigure()
+        except Exception as e:
```
NucciTheBoss: Surprised that the linter/typechecker didn't attempt to fry your chicken strips for using `Exception` here. This is generally a frowned upon practice because it makes the exception handler too broad. We should instead try to unify all the exceptions we know can be emitted here under `SlurmOpsError` - i.e. update `_get_new_entry` to raise `SlurmOpsError` instead of `ValueError`, or update this `try`/`except` block to explicitly catch both errors:

 
```suggestion
        except (SlurmOpsError, ValueError) as e:
```

My preference would be to just unify everything under `SlurmOpsError`, but you are using `ValueError` for its intended purpose in `_get_new_entry`, so I'll default to what you think is best here. The point still stands however that we should be explicit about what types of errors we expect to handle here.

charms/slurmctld/src/state.py
```diff
@@ -151,6 +152,25 @@ def slurmctld_ready(charm: "SlurmctldCharm") -> bool:
...
+    except SlurmOpsError as e:
+        return ConditionEvaluation(False, e.message)
+
+    num_keys = len(keys.get("keys", []))
+    if num_keys == 0:
+        return ConditionEvaluation(False, "Authentication key file contains no keys")
+    if num_keys > 1:
+        return ConditionEvaluation(
+            False,
+            "Authentication key rotation in progress. Waiting for rotation to complete.",
```
NucciTheBoss: [nit]: Generally we don't end status messages with a period.

```suggestion
            "Authentication key rotation in progress. Waiting for rotation to complete",
```

charms/slurmctld/src/state.py
```diff
@@ -20,6 +20,7 @@
 import ops
 from charmed_hpc_libs.ops.conditions import ConditionEvaluation
 from constants import HA_MOUNT_INTEGRATION_NAME
+from slurm_ops.core.errors import SlurmOpsError
```
NucciTheBoss: You can import `SlurmOpsError` directly as it's re-exported in the top-level *\_\_init\_\_.py* file in `slurm-ops`:


```suggestion
from slurm_ops import SlurmOpsError
```

pkg/slurmctld-interface/src/charmed_slurm_slurmctld_interface/__init__.py
```diff
@@ -39,7 +40,9 @@
 
 _logger = logging.getLogger(__name__)
 
-AUTH_KEY_TEMPLATE_LABEL = Template("integration-$id-auth-key-secret")
+AUTH_KEY_LABEL = "auth-key-secret"
```
NucciTheBoss: This label name is a bit too generic. I'd be worried about accidental collisions with other applications that use the same label name. How about `slurm-auth-key-secret` instead?


```suggestion
AUTH_KEY_LABEL = "slurm-auth-key-secret"
```

We could probably drop the `-secret` from the end of the label name because we could likely assume that `slurm-auth-key` corresponds to a secret if we see it in the output of `juju secrets`.

NucciTheBoss (changes requested):
Couple of suggestions and nits around implementation details. Overall, it's looking quite good. I have some suggestions around how we could streamline the key generation API and ensure that experiences are consistent across the charms.

Outside of that, I see that a lot of TODO comments have been added. We should get those converted into issues otherwise they'll get "lost in the sauce", and we'll forget to prioritize them when fixing bugs next cycle. Also, side tangent, but we should come up with a standard format for `TODO`/`FIXME` comments in our code. PyCharm has a useful tool for jumping quickly between `TODO` comments, but you need to format your messages a certain way. Something to standardize on after we complete the rest of the implementation work we have to do this cycle. Probably a good 15 minute discussion after lunch at engineering sprint when our brains are still recovering from all the food 🌶️ 

charms/slurmctld/src/charm.py
dsloanm: It's a Secrets limitation. Keys must match [this regex](https://github.com/canonical/operator/blob/main/ops/model.py#L1303) otherwise:

```
ValueError: Invalid secret keys: ['key_id']. Keys should be lowercase letters and digits, at least 3 characters long, start with a letter, and not start or end with a hyphen.
```

internal/slurm-ops/src/slurm_ops/core/base.py
dsloanm: Updated to also include the UUID per another comment

tests/integration/test_charm.py
dsloanm: No `sinfo` or `sacct` equivalent for slurmrestd and it wasn't clear if `curl`-ing the API would test the auth key or JWT key. My plan is to revisit this during JWT key rotation testing.

charms/slurmctld/src/charm.py
dsloanm: Holdover from testing - fixed

internal/slurm-ops/src/slurm_ops/core/base.py
dsloanm: Changed to `See Also`.

"Notes" or "Note" doesn't appear specified at https://google.github.io/styleguide/pyguide.html. These examples https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html use "Note" but may not be authoritative.

charms/sackd/src/charm.py
dsloanm: Holdover from when this block was copied over multiple charms. Fixed

charms/sackd/src/charm.py
dsloanm: Fixed to `logger.error`. Both log at ERROR level but previous iterations of this block did not include `e` explicitly so `logger.exception` was used

dsloanm:
Integration tests failed with:  `.apps['login'].app_status.message = 'hook failed: "install"'`. Will rerun later.

I suspect the error is related to infrastructure issues and not with the PR code. The same error has been seen in other PRs and no changes are made to `sackd`'s install hook here. The tests succeed locally.

dsloanm:
Review comments addressed. New issues:

* https://github.com/charmed-hpc/slurm-charms/issues/210
* https://github.com/charmed-hpc/slurm-charms/issues/209
* https://github.com/charmed-hpc/slurm-charms/issues/208
* https://github.com/charmed-hpc/hpc-libs/issues/126 (opened by @NucciTheBoss)

Ready for another look

tests/integration/test_charm.py
NucciTheBoss: Ack - sounds like a good plan to me. Yeh, my expectation would be that we need to use `curl` here to exercise the API and get JSONable responses.

internal/slurm-ops/src/slurm_ops/core/base.py
```diff
@@ -501,34 +504,160 @@ def path(self) -> Path:
...
     @property
     def path(self) -> Path:
-        """Get the path to the `slurm.key` secret file."""
+        """Get the path to the `slurm.jwks` secret file."""
         return self._file
 
+    def _write(self, data: dict[str, list[dict[str, str]]]) -> None:
+        self._file.write_text(json.dumps(data))
+        self._file.chmod(0o600)
+        shutil.chown(self._file, self._user, self._group)
```
NucciTheBoss, line 634: [nit]: should this method have docstring? `ruff` doesn't have a hard and fast rule that requires protected methods to have docstrings, but `_write` does look a little out of place without one. We should also probably clarify that data should be the _slurm.jwks_ file data:

```suggestion
    def _write(self, data: dict[str, list[dict[str, str]]]) -> None:
        """Write out slurm.jwks data to a file.
        
        Args:
            data: slurm.jwks file data to write out.
        """
        self._file.write_text(json.dumps(data))
        self._file.chmod(0o600)
        shutil.chown(self._file, self._user, self._group)
```

charms/slurmctld/src/charm.py
NucciTheBoss: Peak "ihateregex" moment. Thank you for the justification and highlighting the specific bit of code in `ops`. We could go with `key-id` for improved readability, but I'll let you decide if you think it's  worth adding the dash here.

charms/sackd/src/charm.py
```diff
@@ -116,6 +120,43 @@ def _on_slurmctld_disconnected(self, event: SlurmctldDisconnectedEvent) -> None:
...
+                    "Failed to retrieve Slurm authentication key. See `juju debug-log` for details"
+                )
+            )
+
+        self.sackd.key.set(auth_key, auth_key_id)
+
+        # Necessary to load new key from file into the service
+        # TODO: replace with self.service.reload()
+        try:
+            call("/usr/bin/systemctl", "reload", "sackd.service")
```
NucciTheBoss: You can use the `systemctl` function from `charm_hpc_libs` here instead of `call` so you don't have to spell out the full filesystem path to the `systemctl` executable:

```python
from charmed_hpc_libs.ops import systemctl

...

systemctl("reload", "sackd")
```

You also don't need to use the full `sackd.service` unit name here. You can just use `sackd`.

NucciTheBoss (changes requested):
Just a few small nits and then I think we're good here 🤩

charms/sackd/src/charm.py
dsloanm: Switched to `systemctl()` but I've kept `sackd.service` to avoid ambiguity with other systemd unit types. I'm aware `systemctl` defaults to .service if not specified but this makes it explicit. Use of the full filesystem path was intentional to guard against $PATH issues including [path hijacking](https://en.wikipedia.org/wiki/PATH_(variable)#Security_implications)

charms/sackd/src/charm.py
```diff
@@ -116,6 +119,43 @@ def _on_slurmctld_disconnected(self, event: SlurmctldDisconnectedEvent) -> None:
...
+        # Necessary to load new key from file into the service
+        # TODO: replace with self.service.reload()
+        try:
+            systemctl("reload", "sackd.service")
+        except SystemdError as e:
+            logger.error("failed to reload sackd.service. reason:\n%s", e)
+            event.defer()
+            raise StopCharm(
+                ops.BlockedStatus(
+                    "Failed to reload sackd.service. See `juju debug-log` for details"
```
NucciTheBoss: [nit]: Using `sackd.service` over just `sackd` for the systemctl command is justifiable - although I don't fully grasp where the ambiguity about the systemd unit type would stem from as in my own experience I've never seen someone fully spell out .service at the end -  but that's my own anecdotal experience, explicit is better than implicit, and it's functionally the same.

That said, this blocked status message deviates from the format we have been using. For example, if we fail to apply an updated configuration to the `slurmd` charm, this is the blocked status message that gets set: 

https://github.com/charmed-hpc/slurm-charms/blob/01f86dfbfbcd1220f5545f295bf5548957e3ac58/charms/slurmd/src/charm.py#L171-L178

Here we're referring to sackd systemd unit name verbatim. I'd prefer if we kept with the current format that we're using, and whoever is tasked with triaging the blocked status message can see the exact systemctl command that was called in the output of `juju debug-log`.

Here's an updated message that is in line with our current message format, but I wasn't sure if we should use configuration or service since we're calling `reload` on the service but effectively we're just directing sackd to reload it's configuration:

```suggestion
                    "Failed to reload `sackd` configuration. See `juju debug-log` for details"
```

NucciTheBoss (changes requested):
One final nit about the status message format in the sackd charm and then I think I'm good to go 👍

NucciTheBoss (approved):
LGTM! :shipit: 

Nice work on this @dsloanm 🚀

PR #213, opened 2026-04-22
feat(sackd/slurmd): add `metrics-endpoint` integration

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

This PR adds the metrics endpoint integration to the `sackd` and `slurmd` charms. Key changes include:

1. Adding support for the `prometheus_scrape` interface to both charms.
2. Adding a `node_exporter` property to the `SlurmManager` base class. I made this a cached property because we don't need to create a new `NodeExporterManager` object each time we access the property. The object will be created once and then reused each time the property is accessed.
3. Adding a `relabel_config` for the metrics exported by `node-exporter`. It takes the `juju_unit` label added by the `MetricsEndpointProvider` class and formats it to the node name format used by our charms. The regex'ing is a built-in mechanism of Prometheus, not something I came up with myself 😂
4. Adding a target configuration that ensures Prometheus/OpenTelemetry Collector will scrape all slurmd/sackd units every 60 seconds.
5. Enabling the `systemd` collector so that we can determine the health of all the Slurm services on each node.

- Related to our ongoing work to add alerts for controllers and compute nodes to Charmed HPC.

## Docs

Or:

I'll be batching all my documentation changes together.


charms/sackd/src/charm.py
```diff
@@ -57,6 +67,30 @@ def __init__(self, framework: ops.Framework) -> None:
...
+                    "static_configs": [
+                        {"targets": [f"*:{NODE_EXPORTER_PORT}"]},
+                    ],
+                    "relabel_configs": [
+                        {
+                            "source_labels": ["juju_unit"],
+                            "target_label": "node",
+                            "regex": r"(\S+)\/(\d+)",  # Match unit name.
+                            "replacement": "${1}-${2}",
+                        }
```
jedel1043: Can you document why we need to apply this replacement? It doesn't seem very obvious at first glance, so it would be good to have a rationale for it.

charms/slurmd/src/charm.py
```diff
@@ -103,33 +134,71 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+            self.slurmd.node_exporter.set_collectors(NODE_EXPORTER_COLLECTORS)
+            self.slurmd.node_exporter.service.enable()
+        except SnapError as e:
+            logger.error(e.message)
+            event.defer()
+            raise StopCharm(
+                ops.BlockedStatus(
+                    "Failed to install `node-exporter`. See `juju debug-log` for details"
+                )
+            )
```
jedel1043: Hmmm I don't love that we're duplicating so much of the same code here. Maybe it would be better to have a map of class objects to error messages?

```python
ERROR_MSG = {
    SlurmOpsError: "Failed to install `slurmd`",
    SnapError: "Failed to install `node-exporter`",
    ...
}
# ...
except (SlurmOpsError, SnapError, ...) as e:
    msg = ERROR_MSG(type(e))
    logger.error(e.message)
    event.defer()
    raise StopCharm(ops.BlockedStatus(f"{msg}. See `juju debug-log` for details"))
```

jedel1043: Alternatively we could use match case or a protocol class, but just having a map seemed like a cleaner option.

jedel1043 (dismissed):
Looks great! I have some small suggestions

charms/sackd/src/charm.py
```diff
@@ -57,6 +67,30 @@ def __init__(self, framework: ops.Framework) -> None:
...
+            PROMETHEUS_SCRAPE_INTEGRATION_NAME,
+            jobs=[
+                {
+                    "job_name": f"juju_{topology.identifier}_node-exporter",
+                    "metrics_path": NODE_EXPORTER_METRICS_PATH,
+                    "scrape_interval": NODE_EXPORTER_SCRAPE_INTERVAL,
+                    "static_configs": [
+                        {"targets": [f"*:{NODE_EXPORTER_PORT}"]},
+                    ],
+                    "relabel_configs": [
```
dsloanm: `MetricsEndpointProvider` docs give these settings as ["strongly discouraged"](https://github.com/canonical/prometheus-k8s-operator/blob/da0b3c7/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L172) and state ["Virtually no charms should use these settings"](https://github.com/canonical/prometheus-k8s-operator/blob/da0b3c7/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L189). Is there an alternative approach? If not, we should provide justification.

charms/sackd/src/charm.py
```diff
@@ -57,6 +67,30 @@ def __init__(self, framework: ops.Framework) -> None:
             self._on_slurmctld_disconnected,
         )
 
+        topology = JujuTopology.from_charm(self)
+        self.metrics_endpoint = MetricsEndpointProvider(
+            self,
+            PROMETHEUS_SCRAPE_INTEGRATION_NAME,
+            jobs=[
+                {
+                    "job_name": f"juju_{topology.identifier}_node-exporter",
```
dsloanm: `{topology.identifier}` may cause issues here. `MetricsEndpointProvider` docs state:

[`job_name` should be a fixed string (e.g. hardcoded literal) [...] if you include variable elements, like your `unit.name`, it may break the continuity of the metrics time series](https://github.com/canonical/prometheus-k8s-operator/blob/da0b3c7/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L167)

dsloanm (changes requested):
Comments inline

charms/slurmd/charmcraft.yaml
```diff
@@ -48,12 +48,14 @@ parts:
...
+    description: |
+      Provides application metrics to COS Prometheus instance.
+    interface: prometheus_scrape
+    optional: true
   slurmctld:
     interface: slurmd
     limit: 1
-  cos-agent:
-    interface: cos_agent
-    limit: 1
```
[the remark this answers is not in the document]
NucciTheBoss: This can be ignored since we haven't cut our first stable release yet. We really don't want to use the `cos-agent` interface since the OpenTelemetry Collector operator doesn't scale well because of its peer integration.

charms/slurmd/src/charm.py
NucciTheBoss: Eh, I think the `dispatch` is the better call here. `match`/`case` is too much indentation for me and using a protocol would likely require re-examining how we raise/emit errors in all our code since the `Error` class from `charmed-hpc-libs` is rather foundational.

charms/sackd/src/charm.py
```diff
@@ -57,6 +67,30 @@ def __init__(self, framework: ops.Framework) -> None:
...
+                        {
+                            "source_labels": ["juju_unit"],
+                            "target_label": "node",
+                            "regex": r"(\S+)\/(\d+)",  # Match unit name.
+                            "replacement": "${1}-${2}",
+                        }
+                    ],
+                }
+            ],
+        )
```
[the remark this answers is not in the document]
NucciTheBoss: Since we're not handling any events related to the `metrics-endpoint` integration, the existing unit tests already cover this functionality. The `prometheus_scrape` charm library contains the unit tests necessary for validating that the `MetricsEndpointProvider` class works as expected.

charms/slurmd/src/charm.py
```diff
@@ -89,7 +98,29 @@ def __init__(self, framework: ops.Framework) -> None:
...
+                        {
+                            "source_labels": ["juju_unit"],
+                            "target_label": "node",
+                            "regex": r"(\S+)\/(\d+)",  # Match unit name.
+                            "replacement": "${1}-${2}",
+                        }
+                    ],
+                }
+            ],
+        )
```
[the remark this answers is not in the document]
NucciTheBoss: Same as before - this functionality is covered by existing unit tests since we're not handling any events related to the `metrics-endpoint` integration. The `MetricsEndpointProvider` class is already tested by the `prometheus_scrape` charm library.

internal/slurm-ops/src/slurm_ops/core/constants.py
```diff
@@ -14,6 +14,12 @@
 
 """Constants used within the `slurm-ops` package."""
 
+NODE_EXPORTER_COLLECTORS = {"systemd"}
+NODE_EXPORTER_METRICS_PATH = "/metrics"
+NODE_EXPORTER_PLUGS = {"hardware-observe", "mount-observe", "network-observe", "system-observe"}
```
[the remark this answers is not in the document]
NucciTheBoss: This is a nit since the order in which the interfaces are connected isn't important, but yeh determinism is cool...

charms/slurmd/src/charm.py
```diff
@@ -103,33 +134,71 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+            )
+
+        self.unit.status = ops.MaintenanceStatus("Installing `node-exporter`")
+        try:
+            self.slurmd.node_exporter.install()
+            for plug in NODE_EXPORTER_PLUGS:
+                self.slurmd.node_exporter.connect(plug)
+            self.slurmd.node_exporter.set_web_listen_address(f":{NODE_EXPORTER_PORT}")
+            self.slurmd.node_exporter.set_collectors(NODE_EXPORTER_COLLECTORS)
+            self.slurmd.node_exporter.service.enable()
```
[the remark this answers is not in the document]
NucciTheBoss: Nah, it works fam. The port just isn't exposed publicly to applications outside the model. The metrics are scraped by the OpenTelemetry Collector application in the same model and are then forwarded to the Prometheus server outside the model.

charms/sackd/src/charm.py
```diff
@@ -79,7 +114,22 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
             )
 
-        self.unit.open_port("tcp", SACKD_PORT)
+        self.unit.status = ops.MaintenanceStatus("Installing `node-exporter`")
+        try:
+            self.sackd.node_exporter.install()
+            for plug in NODE_EXPORTER_PLUGS:
+                self.sackd.node_exporter.connect(plug)
+            self.sackd.node_exporter.set_web_listen_address(f":{NODE_EXPORTER_PORT}")
+            self.sackd.node_exporter.set_collectors(NODE_EXPORTER_COLLECTORS)
```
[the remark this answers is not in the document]
NucciTheBoss: Opening the port isn't necessary. See my earlier comment for additional context.

charms/sackd/src/charm.py
NucciTheBoss: Sure, I might change this value to a constant anyhow to reduce boilerplate. There I can offer additional context that this relabel config is used to make it easier to match node-exporter metrics with metrics collected from the Slurm exporter.

charms/sackd/src/charm.py
NucciTheBoss: I'll add a comment. The alternative is to use the Prometheus Scrape Config K8s operator, but I really dislike the idea of having to people deploy an entirely separate operator and load a very specific YAML file just to get working alerts from the Slurm charms.

Basically they just really don't want charms surfacing these options through `juju config` because the likelihood of someone entering a bad configuration is quite high. We're using constants instead which is safer. My gut feeling is that the library assumes that the exporter already provides the necessary labels for crafting alerts, but Slurm doesn't actually do that.

charms/sackd/src/charm.py
NucciTheBoss: I'll look into this. `topology.identifier` might not be necessary here since `MetricsEndpointProvider` overwrites quite a lot of job information.

internal/slurm-ops/src/slurm_ops/core/constants.py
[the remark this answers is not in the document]
NucciTheBoss: Switch to using lists since mutability is also pretty darn cool.

charms/sackd/src/charm.py
NucciTheBoss: Can confirm that `topology.identifier` isn't actually required here. The `prometheus_scrape` library will acutally overwrite the static name with `topology.identifier`. Thanks for calling this out!

charms/slurmd/src/charm.py
NucciTheBoss: I modified the blocks of code here to use a dispatch to determine the status message that should be surfaced to the user. We might have to do something clever once we add support for the DCGM exporter to ensure we can get unique error messages for both exporters, but for now we could likely wrap the dcgm exporter install under `gpu.autoinstall` (maybe)

charms/sackd/src/charm.py
NucciTheBoss: Added a comment in `slurm-ops`'s constants.py module.

charms/sackd/src/charm.py
NucciTheBoss: I added a comment in `slurm-ops`'s constants.py module.

NucciTheBoss:
@dsloanm @jedel1043 ready for round 2!

I've addressed both of your initial batch of comments. I deduplicated the error handling code in both `sackd` and `slurmd` - but still attempted to ensure that we have granular status messages - and I added a comment to the Prometheus scrape config explaining why we need a relabel config. It also turned out that I didn't need the `topology.identifier` bit of code - the `prometheus_scrape` library handles this for us - so I removed that too. I also centralized the scrape config for the `node-exporter` service in `slurm-ops`, so there's less new constants as well.

dsloanm (approved):
Looks good with latest changes. Test failure seems unrelated to this PR

NucciTheBoss:
I agree that the CI failure is sus given that tests are passing locally. Browsing the logs shows that the failure was caused by the install hook for the slurmctld charm failing, but it works in my own local VM :shrug:

PR #231, opened 2026-05-22
feat: build Slurm charms with the `uv` plugin instead of the `charm` plugin

### LINKED ISSUE #200: Migrate to using the `uv` plugin to build the Slurm charms
The latest CI errors are due to a mismatch between the pre-built version of a Rust-based Python package we are downloading, and the new, updated version that `charmcraft` is attempting to pull in during build time.

The root cause of the error is that `setuptools` is trying to build the Rust-based Python packages from source with the `rustc` compiler which isn't available in the `charmcraft` runtime environment. The `charm` plugin mandates that only sdist-only builds are used, so we use clever magic in the background to seed `charm-binary-python-packages`. It's really not ideal to build packages from source with `rustc` because it is very resource intensive.

To circumvent this issue once and for all, we can use the `uv` plugin instead since it allows us to pull in wheels rather than sdist-only builds. It will also help us avoid needing magic to cache the version of binary Python packages we want in _charmcraft.yaml_.

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

This PR updates our build tooling so that the Slurm charms are built using the `uv` plugin instead of the "deprecated" `charm` plugin. Key changes include:

1. Updating each of the charms' _charmcraft.yaml_ files to use the `uv` plugin.
2. Updating each of the workspace _pyproject.toml_ files to explicitly declare a charms' required dependencies. There were several places where we were implicitly pulling in dependencies which made the `uv` plugin unhappy.
3. Manually declaring `python-apt` as a dependency. The new way charmcraft packs charms doesn't allow dependencies from outside a packed charms' virtual environment to be used, so this was breaking the `slurmd` charm's `gpu` module when it attempted to import the `apt_pkg` model.
4. Dropping several dependencies like `dbus-fast` and `pycryptodome` which are no longer used in any of the Slurm charms.
5. Patching _repository.py_ to inject local dependencies into _pyproject.toml_ files. This is required to provide local packages to `uv` since it doesn't operate on _requirements.txt_ files. This change required introducing `rtoml` as a development dependency since Python's built-in `tomllib` doesn't have writing capabilities.
6. Dropping `charm-binary-python-packages` completely. This option is no longer required since `uv` is capable of pulling prebuilt wheels. No more fancy magic with _uv.lock_.

- Closes https://github.com/charmed-hpc/slurm-charms/issues/200

## Docs

Or:

This is a non-user facing build tooling change.

### DIFF
--- a/charms/sackd/charmcraft.yaml
+++ b/charms/sackd/charmcraft.yaml
@@ -36,8 +36,12 @@ parts:
     source: .
     build-snaps:
       - astral-uv
+    override-build: |
+      # Regenerate lockfile to use vendored subpackages from `slurm-charms`.
+      uv lock
+      craftctl default
     build-environment:
-      - UV_FROZEN: "false"  # Generate charm lockfile at build time.
+      - UV_FROZEN: "false"  # Allow lockfile to be updated in `override-build`.
 
 provides:
   metrics-endpoint:
--- a/charms/sackd/charmcraft.yaml
+++ b/charms/sackd/charmcraft.yaml
@@ -32,8 +32,12 @@ platforms:
 
 parts:
   charm:
-    build-packages:
-      - git
+    plugin: uv
+    source: .
+    build-snaps:
+      - astral-uv
+    build-environment:
+      - UV_FROZEN: "false"  # Generate charm lockfile at build time.
 
 provides:
   metrics-endpoint:
--- a/charms/sackd/pyproject.toml
+++ b/charms/sackd/pyproject.toml
@@ -3,9 +3,9 @@ name = "sackd"
 version = "0.0"
 requires-python = "==3.12.*"
 dependencies = [
-    "ops",
+    "ops ~= 3.0",
     "charmed-hpc-libs[ops] >= 0.1.0",
-    "cosl",
+    "cosl ~= 1.0",
     "charmed-slurm-sackd-interface",
     "slurm-ops"
 ]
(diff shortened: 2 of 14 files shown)


charms/sackd/charmcraft.yaml
```diff
@@ -32,8 +32,12 @@ platforms:
...
 parts:
   charm:
-    build-packages:
-      - git
+    plugin: uv
+    source: .
+    build-snaps:
+      - astral-uv
+    build-environment:
+      - UV_FROZEN: "false"  # Generate charm lockfile at build time.
```
jedel1043: For this, maybe it would be better to copy the lock file from the root directory instead? That way dependencies cannot differ between the root uv.lock and the lock files generated by charmcraft.

NucciTheBoss: I can try injecting the root _uv.lock_ file and see what happens.

NucciTheBoss: Updated the implementation to now inject the root uv lockfile.

NucciTheBoss:
@jedel1043 ready for review round 2.

I updated the _repository.py_ implementation so that the top-level _uv.lock_ file is injected into the stage charm directory. I still have to do a bit of magic in the `charmcraft` build environment since the lockfile must be updated to successfully capture the vendored packages under `dist/`, but the regenerated lockfile will use the existing dependency versions in the top-level lockfile.

jedel1043 (approved):
Looks great! And thankfully the hack to update the lock file with our vendored packages is very small.

PR #269, opened 2026-06-26
feat: add monorepo package publishing workflow

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

Add a GitHub Actions workflow that calls the monorepo package publishing callable workflow from `canonical/hpc-team`. The workflow publishes packages from the `pkg/` directory to PyPI when their version is bumped on `main`, per UHPC009.

Closes HPCENG-875.

## Docs

### DIFF
--- a/.github/workflows/publish-packages.yaml
+++ b/.github/workflows/publish-packages.yaml
@@ -1,0 +1,26 @@
+# Copyright 2026 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-packages.yaml
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+
+name: Publish Python packages
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "pkg/*/pyproject.toml"
```
jedel1043, line 22: Oh, this makes it such that this runs only if we bump the version in the pyproject.toml, right? That's nice!

jedel1043 (approved):
Good thing that we could just reuse our shared actions after the migration.

.github/workflows/publish-packages.yaml
NucciTheBoss: Yep, this workflow will only trigger if we update the version of a public package under `pkg`. It'll just require calling `uv version ...` to bump the version of the package in the _pyproject.toml_ file. The workflow will also check if the modification was actually a version bump or if we just changed the packages' workspace configuration.

PR #273, opened 2026-07-23
feat: migrate charms to Ubuntu 26.04

jedel1043:
# Pre-submission checklist

## Summary of changes

Migrates all the Slurm charms to use base 26.04, and migrates all our actions to use 26.04.
Also adds a dependabot.yaml file (didn't want to open a new PR for it to save some CI time).

## Docs

* [x] I have created a pull request to add or update relevant documentation in [canonical/charmed-hpc-docs](https://github.com/canonical/charmed-hpc-docs) or another documentation location.

Created https://github.com/canonical/charmed-hpc-docs/pull/158.


charms/slurmctld/tests/unit/test_charm.py
NucciTheBoss, line 1: Why did you have to add additional mocks here? Is there a compatibility issue somewhere, or is this a required changed introduce with upgrading to a new `charmlibs-apt` version?

charms/slurmctld/tests/unit/test_ha.py
```diff
@@ -79,6 +79,7 @@ def test_ha_migration_empty_filesystem(
             capture_output=True,
             text=True,
             check=True,
+            env=None
```
NucciTheBoss: Why did you need to add `env=None` here? We're there some conflicts in your environment that made it preferable to not inherit the outer scope's environment variables?

repository.py
```diff
@@ -375,15 +394,21 @@ def stage_charm(
...
+        dest = charm.build_path / "lib" / "charms" / lib.path
+        logger.debug("Copying %s to %s", lib, dest)
+        if not dry_run:
+            copy(src, dest)
 
-        with open(charm.build_path / "pyproject.toml", "r") as fin:
+    if not dry_run:
+        with open(charm.build_path / PYPROJECT_FILE, "r") as fin:
             pyproject = rtoml.load(fin)
+        # Ensure `tool.uv.sources` and `tool.uv.dependency-metadata` always exist
```
NucciTheBoss, line 408: [nit]: Generally I like to have blank line after a `with` block for easier visual separation. Makes it easier for me to interpret a context manager independently from the outer scope 😅


```suggestion
        
        # Ensure `tool.uv.sources` and `tool.uv.dependency-metadata` always exist
```

charms/slurmctld/src/constants.py
```diff
@@ -65,3 +65,37 @@
...
+gM+hv42vPTOnCo+apD97/VArsp59pDqEVoAtMTk72fdBqR+BB77g2hBkKESgQIEq
+EiE1/TOISioMkE0AuUdaJ2ebyQXugSHHuBaqbEC47v8t5DVN5Qr9OriuzCuSDNFn
+6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
+IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
+ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
+fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
+=cs1s
+-----END PGP PUBLIC KEY BLOCK-----
+"""
+UBUNTU_HPC_SLURM_PPA_URI = "https://ppa.launchpadcontent.net/ubuntu-hpc/slurm-wlm-25.11/ubuntu"
```
NucciTheBoss: Won't enabling from the 25.11 PPA here cause the `slurmctld` charm to pull the `slurmctld`, `slurm-client`, etc packages from the Ubuntu HPC PPA and not the Universe repo? If so, this means that all the other Slurm charms will be pulling a different `slurm-wlm` packageset while this charm still pulls from the PPA.

The integration tests are still passing, but I can see this being an issue in the future if our Slurm PPA gets behind on security updates that are being rolled out to Universe. It might be better to create a dedicated PPA for the `slurm-mail` package so that we can still pull this out-of-distro package, but ensure that all the Slurm charms are still pulling the same `slurm-wlm` package set from Ubuntu Universe. You can just copy the existing `slurm-mail` Debian packages from the slurm-wlm-25.11 PPA to dedicated PPA for `slurm-mail`.

jedel1043: In this case it should be fine because our PPA is copying the Resolute version directly from archive. However, it does sound better to have `slurm-mail` on its own PPA, so I'll do that.

charms/slurmctld/tests/unit/test_charm.py
jedel1043: Because I moved the slurm-mail PPA setup from slurm-ops to slurmctld, since we don't require it for all charms anymore. This'll change when I setup the new slurm-mail PPA, but we will still require the patches

charms/slurmctld/tests/unit/test_ha.py
jedel1043: This is not a call to `rsync`, this asserts that the code calls `subprocess.run` with the specified arguments. `call` comes from `unittest.mock`.

I didn't investigate it in detail since it was a trivial fix, but I assumed it was a change in behaviour in `unittest.mock`, where it previously ignored `None` arguments for equality with mocks but not anymore.

charms/slurmctld/src/constants.py
jedel1043: Fixed in f39e947

charms/slurmctld/src/mail.py
```diff
@@ -135,8 +142,22 @@ def install() -> None:
...
+            enabled=True,
+            repotype="deb",
+            uri=UBUNTU_HPC_SLURM_MAIL_PPA_URI,
+            release=distro.codename(),
+            groups=["main"],
+        )
+        ppa.import_key(UBUNTU_HPC_PPA_KEY)
+        repositories = apt.RepositoryMapping()
+        repositories.add(ppa)
+        apt.update()
```
NucciTheBoss, line 158: [foresight]: Eventually I want us to upstream PPA configuration into `charmed-hpc-libs` so that we don't have to copy the same bit of bolierplate for each charm we have that configures an external repository on Launchpad. I'm think of naming this common method `add_ppa(...)` or something similar on `AptOpsManager`.

jedel1043: Agreed, since that's the only missing piece we need to migrate away from charmlibs-apt

charms/slurmctld/tests/unit/test_ha.py
NucciTheBoss: I wanted a better understanding of why this change was necessary, but it's also not major enough to warrant a deeper amount of time speculating on why it's necessary. Closing this conversation as resolved.

PR #275, opened 2026-08-04
refactor(slurm-ops): drop distro in favour of platform

### LINKED ISSUE #256: Consider using Python standard `platform` instead of `distro`
The Slurm charms depend on [`distro`](https://pypi.org/project/distro/) to provide the distribution code name ("jammy", "noble") to set up the Ubuntu HPC PPA:

https://github.com/charmed-hpc/slurm-charms/blob/c47c80c36a2479eb897f732db7f8d34c2b613e50/internal/slurm-ops/src/slurm_ops/core/base.py#L204-L210

Python 3.10 added  [`platform.freedesktop_os_release()`](https://docs.python.org/3/library/platform.html#platform.freedesktop_os_release), providing standard access to the `/etc/os-release` file. Now, `platform.freedesktop_os_release().get("VERSION_CODENAME")` can provide the code name.

We should consider using this over `distro` to reduce our set of external dependencies. There are no other uses of `distro` in the code and we do not support Python < 3.10 (`pyproject.toml` specifies `requires-python = "==3.12.*"`).

Krishnanand-G:
## Summary
- Replace the `distro` dependency with `platform.freedesktop_os_release()` when resolving the Ubuntu codename for the HPC PPA.
- Drop `distro` from package metadata and the lockfile.
- Update the unit test mock accordingly. `lsb_release` is no longer involved, so the assertion is simpler.

Closes #256

## Test plan
- [x] Unit test `test_init_ubuntu_hpc_ppa` passes
- [x] Confirm no remaining `distro` imports under `internal/slurm-ops`


NucciTheBoss:
Hi there @Krishnanand-G - thank you for taking the time to contribute to Charmed HPC.

I've approved the CI run to validate your changes. In the meantime, there are some merge conflicts due to recent dependency updates, so could you please resolve them before we formally review this pull request?

Krishnanand-G:
Rebased onto main and resolved the conflicts.

The PPA init code had moved into `charms/slurmctld/src/mail.py`, so the `platform.freedesktop_os_release()` swap is there now, and `distro` is gone from the package metadata.

NucciTheBoss:
> The PPA init code had moved into `charms/slurmctld/src/mail.py`, so the `platform.freedesktop_os_release()` swap is there now, and `distro` is gone from the package metadata.

Yep, Slurm 25.11 is included in Ubuntu 26.04, so we no longer need to enable our special Slurm PPA for all the Slurm charms. That said, we still need to pull `slurm-mail` in since we haven't had the chance to upstream it into Debian/Ubuntu universe yet.

Looking at CI, it appears that some unit test mocks that need to be updated. You can run the unit tests and triage the issue with the command `just unit slurmctld`.

Krishnanand-G:
The SMTP install-failure test now stubs `platform.freedesktop_os_release` the same way the success test already did. That was the `/etc/os-release` FileNotFoundError after the rebase.

NucciTheBoss (approved):
LGTM

